PR #8, opened 2020-07-03
build.sh: add the wifi feed to default builds

blogic:
This patch selects the wifi feed when building images in CI.
This has the effect, that backports, haps, wpas, iw, ath10k-ct will the
v5.7 version from current OpenWrt HEAD.

Signed-off-by: John Crispin <john@phrozen.org>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -39,7 +39,7 @@ AP2220)
 	;;
 esac
 cd ${BUILD_DIR}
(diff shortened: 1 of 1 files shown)


greearb:
I previously did a successful stability & performance regression run on 5.7 backports, so assume this is good for pulling.

PR #13, opened 2020-07-14
Remove wpad-basic from build

ArifAlam:
Remove wpad-basic from build


blogic:
I fixed this issue in trunk in a different way, but this one is actually way better so I'll merge this.

patches/pending/0003-profiles-add-the-wlan-ap-board-profiles.patch
```diff
@@ -157,6 +157,7 @@ index 0000000000..b3668bec5d
 +  CONFIG_OPENSSL_WITH_PSK=y
 +  CONFIG_OPENSSL_WITH_SRP=y
 +  CONFIG_OPENSSL_WITH_TLS13=y
++  # CONFIG_PACKAGE_wpad-basic is not set
 +
```
blogic, line 161: merged into pending, Thanks ! (with a minor modification)

PR #15, opened 2020-07-17
802.11v turned on

AmmadRehmatCUS:


blogic:
Hi,
should this not be triggered via the [btm] column ? Also the Patch has some formal problems, such as the SoB line
   John

AmmadRehmatCUS:
> Hi,
> should this not be triggered via the [btm] column ? Also the Patch has some formal problems, such as the SoB line
> John

Hi ,
I am not familiar with SoB line acronym could you please elaborate on that one.

Ammad

blogic:
Hi,
a patch should have this format

---
prefix: short decription

Longer description.

Signed-off-by: Name <email>
---

Also I am not sure that always enabling these 2 options is correct. The VIF table has a BTM collumn. Should this not be used to decide if the feature should be turned on or not ?
  John

AmmadRehmatCUS:
> Hi,
> a patch should have this format
> 
> prefix: short decription
> 
> Longer description.
> ## Signed-off-by: Name
> 
> Also I am not sure that always enabling these 2 options is correct. The VIF table has a BTM collumn. Should this not be used to decide if the feature should be turned on or not ?
> John

Hi John

After consulting with team i have changed the triggering to btm and also SoB has also been added to commit. Please review the PR again.

Ammad

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -388,6 +388,13 @@ bool target_vif_config_set2(const struct schema_Wifi_VIF_Config *vconf,
 		blobmsg_add_bool(&b, "ieee80211r", 0);
 	}
 
+	if (changed->btm) {
+		blobmsg_add_bool(&b, "ieee80211v", 1);
+		blobmsg_add_bool(&b, "bss_transition", 1);
+	} else {
+		blobmsg_add_bool(&b, "ieee80211v", 0);
+	}
+
```
blogic, line 397: The code to write the state back to the cloud was missing in this patch. I added this and added the patch to the pending tree, Thanks !

PR #16, opened 2020-07-20
Staging ssidclient ratelimiting

chaitanyagodavarthi:


blogic:
merged into pending, Thanks !

blogic:
Turns out that the code will write back a state even if no config was set. This was causing cloud-sdk to restart wifi in an infinite loop. I added a patch on top of the PR inside pending to mitigate this. Please have a look to check if i broke anything.

yashvardhan-cu:
Please ignore. Just testing my access

PR #18, opened 2020-07-23
Staging fwupgrade

yashvardhan-netex:
Added support to facilitate firmware upgrade from a tar package for example from the tar.gz files on the jfrog server.
Also included changes to convey the firmware package tar name created in the workflow to the opensync gateway using the ovsdb interface. The gateway would simply use this name against the firmware package name it used initially to trigger the upgrade and determine whether the upgrade was successful once the AP connects back after a reboot.


yashvardhan-netex:
Hi @blogic
I can see that we are using a different workflow for dev builds and are not uploading the dev firmware packages on the Jfrog server.
I have done some changes in the build.yml which I guess by default would run only for trunk builds. I would like to test them. Should I simply do the changes in the  build-dev.yml for the testing.
Can we also upload the dev fw packages to the jfrog artifactory? May be to a new folder other than the trunk.

.github/workflows/build.yml
```diff
@@ -37,8 +38,9 @@ jobs:
           GH_BUILD_USERNAME: ${{ secrets.GH_BUILD_USERNAME }}
           GH_BUILD_PASSWORD: ${{ secrets.GH_BUILD_PASSWORD }}
       run: |
-        make TARGET=ECW5410
         TAG=$(date +%Y-%m-%d)-$(git rev-parse --short HEAD)
+        echo "ecw5410-$TAG" > feeds/wlan-ap/opensync/src/vendor/ecw5410/.pkgname
+        make TARGET=ECW5410
```
blogic, line 43: Should we not just make the opensync Makefile add this ? running make twice seems sub optimal

blogic: sorry, iI did not have enough coffee yet, i missed the prior removal line ;)

yashvardhan-netex: > sorry, iI did not have enough coffee yet, i missed the prior removal line ;)

No problem John :)

blogic:
merged into pending, Thanks !

PR #20, opened 2020-07-28
Staging fixswupgrade

yashvardhan-netex:
Fix for an intermittent upgrade failure issue


blogic:
merged into pending, Thanks !

PR #21, opened 2020-07-29
Staging preservecerts

yashvardhan-netex:
Preserve certificates over FW upgrade


blogic:
OpenWrt provides a feature for this via the Makefiles, I will add it in that manner.

https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=package/system/opkg/Makefile#l60

blogic:
please help test if this achieves the same result -->
https://github.com/Telecominfraproject/wlan-ap/commit/a88b2c03874e34ddecef790ca8abd8dd1b0b97c1

blogic:
@mikehansen1970 confirmed via mail that my patch in pending is working. closing the PR, Thanks !

PR #22, opened 2020-07-29
Staging multissid

RickSommerville:
AP supports multiple SSID (VIF) per radio.
Naming convention is radioX has VIF named wlanX, wlanX-1, wlanX-2, etc.


blogic:
merged into pending, Thanks !

PR #23, opened 2020-07-30
Statging fixmgmtipreporting

yashvardhan-netex:
opensync nm expects exact interface names or else it does not populate
inet address and related parameters in the wifi inet state table


blogic:
Hi,
the file being patched here will actually disappear once we merge the new network manager. the values will then be generated directly from what is present inside UCI
  John

yashvardhan-netex:
> Hi,
> the file being patched here will actually disappear once we merge the new network manager. the values will then be generated directly from what is present inside UCI
> John

Hi John,
That sounds good! Are we planning to introduce the new network manager any time soon or before MVP? If not, Can we have this fix for now and change it later with the new nm?
Yash

blogic:
@yashvardhan-netex I am hoping to merge before mondays meeting

yashvardhan-netex:
> @yashvardhan-netex I am hoping to merge before mondays meeting

Great! I think we can wait till then

PR #24, opened 2020-07-30
Staging client events

AmmadRehmatCUS:


blogic:
Can you please add a proper patch annotation to the patch and the PR. The expected format is

opensync: add $feature

This patch adds xyz, required to make abc work

Signed-off-by: $name <$email>

PR #25, opened 2020-07-31
Staging synthetic client

AmmadRehmatCUS:
synthetic client data path completed


blogic:
I agree with Rick please use the following license header
/* SPDX-License-Identifier: BSD-3-Clause */

blogic:
Can you please add a proper patch annotation to the patch and the PR. The expected format is

opensync: add $feature

This patch adds xyz, required to make abc work

Signed-off-by: $name <$email>

AmmadRehmatCUS:
closing as dns client opened

PR #33, opened 2020-08-03
Staging preserveconfdb

yashvardhan-netex:
Patch to preserve opensync db and schema over firmware upgrade


blogic:
merge into pending, Thanks !

PR #34, opened 2020-08-04
wlan-ap-config: Use qca9984 fw image for ecw5410

ArifAlam:


blogic:
merged into pending, Thanks !

PR #35, opened 2020-08-05
wlan-ap-config: Fix ECW5410/ECW5211 serial and model config

ArifAlam:
- Use eth0 mac address for serial number
- Fix model format


blogic:
merged into pending, Thanks !

PR #37, opened 2020-08-06
netifd: remove beacon_rate param

ArifAlam:
beacon_rate config was introduced as part of rrm commit. Some
drivers don't support this.  Removing this config from
hostapd config.


blogic:
merged into pending, Thanks !

PR #38, opened 2020-08-06
Staging redirectoraddr

yashvardhan-netex:
- Added helper script to apply redirector address
 - Added support to preserve the same


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/utils.c
```diff
@@ -477,3 +477,21 @@ int ieee80211_channel_to_frequency(int chan)
...
 	return 0;
 }
+
+int get_redirector_addr(char *addr)
+{
+	char filepath[PATH_MAX] = {0};
+	char buf[REDIRECTOR_ADDR_SIZE];
+	int fd, ret;
+
+	snprintf(filepath, sizeof(filepath), "/usr/opensync/etc/redirector");
```
blogic, line 487: I would prefer this to be stored in uci, I will prepare an according patch for testing
   John

yashvardhan-netex: Hi John,
It totally makes sense to use a UCI parameter. If you have not already done the changes, should I take it up?

blogic: I have not started yet, so go ahead please

yashvardhan-netex: Sure. I will do that. Thanks!

blogic:
merged into pending, Thanks !

PR #39, opened 2020-08-10
Staging wifi526 ec420 5ghz init

RickSommerville:
EC420 has qca9984 not qca9888.  Adjust firmware config appropriately.


blogic:
merged into pending, Thanks !

PR #40, opened 2020-08-14
WIFI-579 TP Link AP model number display as EC420-G1

RickSommerville:


blogic:
I would propose that fixing this inside entity.c globally might be a better option. allow me to propose a patch to do so start of the week.

RickSommerville:
Sure

PR #41, opened 2020-08-14
WIFI-441 DHCP client support added

AmmadRehmatCUS:
DHCP synthetic client support added

- the dhcp synthetic client sends discovery packet and waits for an offer in order to check whether the server is alive or not.


blogic:
the SoB needs to be the last line of the annotation
also the patch does not add dhcp client support, but a feature that allows use to simulate a client for diag purposes. kindly adapt the description

AmmadRehmatCUS:
> the SoB needs to be the last line of the annotation
> also the patch does not add dhcp client support, but a feature that allows use to simulate a client for diag purposes. kindly adapt the description

@blogic comments and commit messages changed as requested.

blogic:
merged into pending, Thanks !

PR #42, opened 2020-08-17
Staging 80211rbug

AmmadRehmatCUS:
WIFI-560 - FT over DS should be set to zero which wasnt taking effect due to missing declarations.

Signed-off-by: Ammad Rehmat ammad.rehmat@connectus.ai


blogic:
merged into pending, Thanks !

PR #44, opened 2020-08-17
WIFI-628 Adjust wlan naming convention.

RickSommerville:
Replace dashes with underscores to simplify operation with UCI.


blogic:
merged into pending, Thanks !

PR #45, opened 2020-08-19
Staging wifi527 report ssid in client stats

RickSommerville:


blogic:
merged into pending, Thanks !
(added the SoB and fixed the whitespace errors in the patch)

PR #46, opened 2020-08-26
Staging ben

greearb:
I'd like to get this into staging so that I can get ath10k firmware binary crash dumps.


blogic:
having a file in /etc/config/* that is not uci is a no-go I am afraid

blogic:
merged into pending, Thanks !

PR #51, opened 2020-09-04
WIFI-442 : Support for synthetic radius client added

AmmadRehmatCUS:
Support for radius synthetic client added


blogic:
looks good, let the builders run. the profile patch will break as you did not update the diffstat, but i can fix that tomorrow

blogic:
merged into pending, Thanks !

PR #53, opened 2020-09-09
Staging ben

greearb:


greearb:
This pull request updates ath10k-ct firmware and fwcfg file for 4019 radio, this gives more TIDs, at cost of fewer stations, but result should not crash under load due to lack of TID objects.

blogic:
merged into pending, Thanks !

PR #55, opened 2020-09-10
WIFI-705: Set procd to respawn hostapd indefinitely

ArifAlam:
Procd currently only respawns hostapd five times. This commit
will ensure that hostapd is always respawned.


blogic:
merged into pending, Thanks !

PR #57, opened 2020-09-10
WIFI-729 : Command_State now reflects command string as well

AmmadRehmatCUS:
WIFI-729 : Command_State now reflects command string as well


blogic:
merged into pending, Thanks !

PR #63, opened 2020-09-14
Wifi-638 Noise floor inaccurate

RickSommerville:
On-channel survey reports only the active channel.
Off-channel survey reports on channels actually measured.
Channel utilization reported as a percentage.


blogic:
merged into pending, Thanks !

PR #65, opened 2020-09-14
Staging voice vidio detect

chaitanyagodavarthi:


blogic:
Hi,
the nf code is missing the license headers and the osynbc ones have plume and not spdx bsd heders.
   John

feeds/wlan-ap/opensync/patches/18-video-voice-detect.patch
```diff
@@ -3638,7 +3638,6 @@
 +    struct sip_call_start *SipCallStart = NULL;
 +    struct sip_call_end *SipCallEnd = NULL;
 +    struct sip_call_report *SipCallReport = NULL;
-+    int i = 0;
```
blogic: assuming that this fixes code you added prior in the PR it should be folded into the prior patch

chaitanyagodavarthi: should I squash all the patches in to a single commit?

blogic:
merged into pending, Thanks !

PR #67, opened 2020-09-15
opensync: configure ieee80211k

ArifAlam:
Add the ability to enable/disable ieee80211k via custom options


blogic:
merged into pending, Thanks !

PR #70, opened 2020-09-18
OpenWrt build system update

dfranusic:
* **opensync_stats.proto** gets compiled during the build process


blogic:
nice, this has been on my list for a while. we should ask the cloud team to test an image with this patch before we push
@RickSommerville could you help getting aligned with the cloud team on that ?

blogic:
In testing, this patch breaks events for the clouds.

dfranusic:
How do we address this issue? The only connection with the cloud, from our side at least, is the protobuffer definitions (opensync_stats.proto).

RickSommerville:
John:  Which event breaks?  What was tested against?  Did you mean the statistics rather than events?  Or do you mean the build breaks when combined with the earlier pull request for events? 

Damir:  Did you test the various statistics reports against the static gateway with this image?

dfranusic:
Rick: I had no issues with SM reports, as I could clearly see everything being logged in **"opensyncgw.log"**.

blogic:
@RickSommerville You said so when I asked ?!

RickSommerville:
My concern was with the protobuf change, which has already been corrected (by rebasing).

blogic:
ok, let me merge it later today ot tomorrow in that case

blogic:
merged into pending, Thanks !

PR #71, opened 2020-09-21
WIFI-790:Quick Fix for Session Timeout Reporting

NagConnectus:
package:OpenNDS
This patch will resolve the issue:VIF State captive_portal/session_timeout reports 0 value when not set


blogic:
merged into pending, Thanks !

PR #74, opened 2020-09-28
Staging inactive firmware

yashvardhan-netex:
Added support to report inactive firmware version
 - Inactive firmware image version is stored over upgrade
   and is reported in AWLAN_Node table.
 - Rename FW_IMAGE_NAME to FW_IMAGE_ACTIVE

FIXES: WIFI-609


blogic:
merged into pending, Thanks !

PR #75, opened 2020-09-30
ipq807x: Fix dtc host build with GCC 10

ecsv:
The dtc binary has to be compiled using the host GCC and not the target cross compiler GCC. And GCC 10 is using by default -fno-common which inhibits the merging of tentative definitions by the linker.


blogic:
rebased and merged into pending, Thanks !

PR #79, opened 2020-10-07
Fixup ecw5410

greearb:
This fixes ath10k-ct driver/firmware issues on ECW5410-L EU variant.  Root cause was setting regdom to US on this platform causes the radios/firmware to hang hard.  This update no longer sets regdom to anything in fwcfg, leaving system at platform defaults.  Users can override using fwcfg as needed on their platforms of choice.  Tx-buffers are increased from 2200 to 2400 since some extra RAM was available in the firmware.  This should improve throughput with many (100+) stations associated because the firmware can build bigger AMPDUs for the various peers.  This update brings in latest ath10k-ct firmware too.


blogic:
hmkay, lets try and work on those git rebase skills the tree/PR is pretty dirty/tainted :-) manually pulled the 2 commits.

merged into pending, Thanks !

PR #81, opened 2020-10-13
ipq807x: Add batman-adv for wifi-ax's mac80211

ecsv:
The batman-adv source code in openwrt-routing for OpenWrt 19.07 is expecting the backports' compat code from OpenWrt 19.07. But the mac80211 package in wifi-ax is providing the ABI (and API) from a much newer kernel.

To work around this problem, a fork of the batman-adv package from OpenWrt 19.07 without the patches 0001-0004 can be used to switch to the new netlink policy structure from newer kernel versions.


blogic:
Awesome, thanks for taking care of this !

blogic:
merged into pending, Thanks !

PR #92, opened 2020-10-27
Add network manager support and remove dependency on netifd manager

aparna-salian:
Signed-off-by: aparna.salian <aparna.salian@globaledgesoft.com>


blogic:
Hi,
thanks for the patch, there are a few formal errors and if I am reading the code correctly, merging this PR will cause a regression and make ApNos behave differently.
*) please make sure that the patch is properly annotated. the prefix and secription are mssing, the SoB is incorrectly formatted
*) when merging the patch, the ApNos images would be built with NM2 enabled as opposed to the netifd manager.
   John

PR #95, opened 2020-10-29
hostapd: Disconnect client when rssi is low

yashvardhan-netex:
Disconnect client in case rssi drops below the set threshold level

### DIFF
--- a/feeds/wifi-trunk/hostapd/files/hostapd.sh
+++ b/feeds/wifi-trunk/hostapd/files/hostapd.sh
@@ -189,6 +189,9 @@ hostapd_common_add_bss_config() {
 	config_add_int maxassoc max_inactivity
 	config_add_boolean disassoc_low_ack isolate short_preamble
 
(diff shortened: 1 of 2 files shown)


RickSommerville (commented):
At line 31 in ubus.c something happens to the indentation.

In the 810-client-disconnect-... patch, at line 141 the comment says you don't know if this is going to work.  Does it work?

In

yashvardhan-netex:
> At line 31 in ubus.c something happens to the indentation.
> 
> In the 810-client-disconnect-... patch, at line 141 the comment says you don't know if this is going to work. Does it work?
> 
> In

Indentation taken care in the lates commit c316a9b.
Removed comment at line 141 as it does work.

blogic:
Hi few observations
*) the check should not be happening inside ubus.c
2) why is there a ubus method to update the config ? do we want to change this at runtime ?
3) please split the patch up into logical units. there should be 1 part for the hapd changes, one for the ubus part, ...

blogic:
merged into pending, Thanks !

PR #99, opened 2020-11-03
Staging sartura hapd ubus

msenecic:
ubus_collector and hostapd bugfix

* Bugfix in hostapd: fixed improper increment operator usage in os_realloc
  function that caused a crash on multiple SSIDs

* Bugfix in ubus_collector: fixed the improper usage of blobmsg_for_each_attr
  macro and the misplaced evsched_task_reschedule calls

* Reverted the previous client sessions hostapd patch

Signed-off-by: Mario Senecic <mario.senecic@sartura.hr>


blogic:
Hi
please do not revert the revert and then add 3 bug fix patches. rather consider to just add a new patch that includes all 4 with correct annotation and description of the feature
   John

blogic:
merged into pending, Thanks !

PR #100, opened 2020-11-04
wifi-ax: update driver to latest CAF version

blogic:
This depends on QCA pushing latest FW binaries for ath11k to github


blogic:
superseeded by the ED2 staging tree

PR #101, opened 2020-11-04
treewide: add support for owrt 20.x

blogic:
This PR adds support for using the OpenWrt 20.x release as baseline instead of 19.07. both baseline version co-exist in this tree. The selection is triggered by passing --20x to setup.py
This PR will give us
* v5.4 kernel
* containers
* eBPF
* MTK/AX support
* Realtek switch support
* software and HW based flow block offloading
* ...


RickSommerville (commented):
I think this change set should include an update to the README.md file to show the revised forms of the setup procedure and to mention the existence of the wean-switch build.

blogic:
@RickSommerville yup I already have such a change staged locally. I am currently busy testing everything to make sure we dont have any regressions

RickSommerville:
We should update the automated build procedures to build these images.  It should probably put the images in a separate folder and use a different file naming convention to reflect the different kernel.  I saw that the Splash logo will reflect the kernel version - probably need to report from AWLAN_Node or somewhere else as well (if not already).

blogic:
superseeded by the new PR based on the ED2 update

PR #104, opened 2020-11-05
Wifi-1002 correct VIF deletion function.

RickSommerville:
Report VIF state for all wifi-iface in UCI.
Report operational state as up or down.
Use ifname not UCI section name for VIF deletion.


blogic:
merged into pending, Thanks !

PR #112, opened 2020-11-19
Set beacon and multicast tx rates

yashvardhan-netex:
- Support to configure fixed beacon and multicast rates
- Modified/Removed unwanted RRM parameters

### DIFF
--- a/feeds/wlan-ap/opensync/files/etc/hotplug.d/net/50-set-tx-rates
+++ b/feeds/wlan-ap/opensync/files/etc/hotplug.d/net/50-set-tx-rates
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+[ "${ACTION}" = "add" ] || exit 0
+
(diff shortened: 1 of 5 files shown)


RickSommerville (approved):
What happens if the supplied bin_rate or mcast_rate are invalid?  It appears the beacon_code or mcast_code could be written as 0xFF.  Does this have negative consequences?  Would it be better to write nothing at all in this case?  

Never mind checked it myself - OxFF means apply the default behaviour.

yashvardhan-netex:
> What happens if the supplied bin_rate or mcast_rate are invalid? It appears the beacon_code or mcast_code could be written as 0xFF. Does this have negative consequences? Would it be better to write nothing at all in this case?
> 
> Never mind checked it myself - OxFF means apply the default behaviour.

Yes 0xFF applies the driver default.

blogic:
merged into pending, Thanks !

PR #113, opened 2020-11-20
opensync: add support for passpoint icon config

ArifAlam:
- Add support for passpoint icon config


blogic:
merged into pending, Thanks !

PR #116, opened 2020-11-23
treewid: add support for OpenWrt 20.x baseline

blogic:
This PR adds OpenWrt 20.x support in parallel to the 19.07 version. This PR includes the ED2 update.
Signed-off-by: <john@phrozen.org>


blogic:
merged into pending, Thanks !

PR #117, opened 2020-11-23
opensync: add proto definitions for dhcp events

lukakudra:
Signed-off-by: Luka Kudra <luka.kudra@netexperience.com>

### DIFF
--- a/feeds/wlan-ap/opensync/patches/24-dhcp-events-proto-definitions.patch
+++ b/feeds/wlan-ap/opensync/patches/24-dhcp-events-proto-definitions.patch
@@ -2,102 +2,83 @@ Index: opensync-2.0.5.0/interfaces/opensync_stats.proto
 ===================================================================
 --- opensync-2.0.5.0.orig/interfaces/opensync_stats.proto
 +++ opensync-2.0.5.0/interfaces/opensync_stats.proto
-@@ -796,8 +796,98 @@ message EventReport {
+@@ -805,8 +805,79 @@ message EventReport {
          repeated ClientConnectEvent     client_connect_event                = 10;
      }
 
-+    /* DHCP Ack Event */
-+    message DhcpAckEvent {
-+        optional uint32                 x_id                                = 1;
++    /* DHCP Common Data */
++    message DhcpCommonData {
++        required uint32                 x_id                                = 1;
 +        optional uint32                 vlan_id                             = 2;
 +        optional bytes                  dhcp_server_ip                      = 3;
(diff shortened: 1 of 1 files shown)


AmmadRehmatCUS:
@RickSommerville 
@lukakudra 

I would just like to make a recommendation to aggregate all the dhcp events like it is done for client session and into one single dhcp event instead of declaring each type of event under client session. Also I believe addition of timestamps in dhcp events would be useful.

lukakudra:
> @RickSommerville
> @lukakudra
> 
> I would just like to make a recommendation to aggregate all the dhcp events like it is done for client session and into one single dhcp event instead of declaring each type of event under client session. Also I believe addition of timestamps in dhcp events would be useful.

The main reason that we aggregated the client connection events into a client session was the fact that we wanted to generate only one event (client connect event) when the connection was successful and generate all the other events when connection was not successful.
Since we don't correlate dhcp events by session_id or any other equivalent value we decided to declare each dhcp event type separately.

That being said, we could still aggregate them into a single message if we agree that this is the best approach. I think that it would make the code easier to read, but I don't see any other significant benefit.

I agree about the timestamps, they should be added.

lukakudra:
Based on the recommendation from @AmmadRehmatCUS we aggregated all the dhcp events into a single dhcp event and added a timestamp in each dhcp event.

blogic:
It appears that this depends on a patch that is not in the tree ?
/tip/pending$ git am 13c8daaf9db84d5178ef35d652dcdb6738fae8da.patch
Applying: opensync: dhcp events proto update
error: feeds/wlan-ap/opensync/patches/24-dhcp-events-proto-definitions.patch: does not exist in index
Patch failed at 0001 opensync: dhcp events proto update
hint: Use 'git am --show-current-patch' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".

dfranusic:
I'll look into it today and make the necessary adjustments. Thanks.

dfranusic:
It should be fine now. Awaiting your feedback. Thank you.

blogic:
@dfranusic is there already code consuming these data model extensions ?

dfranusic:
@blogic DHCP protobuf models are generated by our StatsManager (@lukakudra) patches, collected and pushed to ubus by the ubus_collector. This should be finished by the end of this week, and we will create a new PR for all of these changes.

blogic:
merged into pending, Thanks !

PR #121, opened 2020-12-02
opensync: Support Mesh configuration

yashvardhan-netex:
- Patch to support basic batman configuraion through ovsdb interface
   in addition to the defaults.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
@@ -39,6 +39,8 @@ extern struct blob_buf del;
 struct blob_buf hs20 = { };
 struct blob_buf osu = { };
 
+struct blob_buf mesh = { };
+
 enum {
 	WIF_ATTR_DEVICE,
 	WIF_ATTR_IFNAME,
@@ -110,6 +112,9 @@ enum {
 	WIF_ATTR_RADIUS_NAS_ID_ATTR,
 	WIF_ATTR_RADIUS_AUTH_REQ_ATTR,
 	WIF_ATTR_RADIUS_ACCT_REQ_ATTR,
+	WIF_ATTR_MESH_ID,
+	WIF_ATTR_MESH_FWDING,
+	WIF_ATTR_MESH_MCAST_RATE,
 	__WIF_ATTR_MAX,
 };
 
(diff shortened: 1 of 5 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -980,119 +1006,172 @@ void vif_hs20_update(struct schema_Hotspot20_Config *hs2conf)
...
+				blobmsg_add_u32(b1, "mtu", 1532);
+			else
+				blobmsg_add_u32(b1, "mtu", atoi(val));
+		}
+	}
+}
+
 bool target_vif_config_set2(const struct schema_Wifi_VIF_Config *vconf,
-			    const struct schema_Wifi_Radio_Config *rconf,
-			    const struct schema_Wifi_Credential_Config *cconfs,
```
blogic, line 985: lots of code reformatting in this patch making it hard to see what has actually changed. could you please remove the reformatting part ?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -980,119 +1006,172 @@ void vif_hs20_update(struct schema_Hotspot20_Config *hs2conf)
...
-		blobmsg_add_string(&b, "ssid", vconf->ssid);
+	vif_config_security_set(&b, vconf);
 
-	if (changed->ssid_broadcast) {
-		if (!strcmp(vconf->ssid_broadcast, "disabled"))
-			blobmsg_add_bool(&b, "hidden", 1);
-		else
-			blobmsg_add_bool(&b, "hidden", 0);
-	}
+	if (!strcmp(vconf->mode, "mesh")) {
```
blogic, line 1064: I am wondering if the code is more readable if we put this and the normal wifi setup each to their own function.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
RickSommerville: I think the whitespace change is to due to the mesh or ap decision resulting in the AP specific parameters adding 1 level of TAB to the whitespace.  The recommendation to separate the mesh and ap portions into two separate helper functions is probably the solution here.

RickSommerville (commented):
Separate the mesh and AP into two separate functions called from target_vif_config_set2().  As John says, the readability will go way up and the changes will be more obvious.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
yashvardhan-netex: Yeah. Already done with the changes. Currently resolving the merge conflicts with your latest commit. New patch should be pushed in a while.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
yashvardhan-netex: Done

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
yashvardhan-netex: This one was badly indented. So had to change.

blogic:
this appears to not apply ontop of latest pending.

yashvardhan-netex:
> this appears to not apply ontop of latest pending.

Rebased.

blogic:
merged into pending, Thanks !

PR #123, opened 2020-12-02
opensync: add various RADIUS configuration

ArifAlam:
- Add functionality to configure RADIUS NAS-Id and Operator-Name
- Add functionality to configure RADIUS accounting interval
- Enable RADIUS Chargeable User Identity

### DIFF
--- a/feeds/wifi-trunk/hostapd/files/hostapd.sh
+++ b/feeds/wifi-trunk/hostapd/files/hostapd.sh
@@ -218,6 +218,8 @@ hostapd_common_add_bss_config() {
 	config_add_string dae_secret
 	config_add_int dae_port
 
+	config_add_boolean request_cui
+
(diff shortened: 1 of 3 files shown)


RickSommerville (commented):
In vif.c @ line 429, what is "126:s"?  Could we use a macro to give it more meaning?  Is it supposed to be 's' or '%s'?
In vif.c @ line 552 when we read it back, the 's' is treated as a variable and loaded into 'type' and then ignored.  The id is used only to validate the data.  You could just make the call sscanf(buf, "126:s:%s",value); to save the extra variables and enforce the format from line 429.  The function should return the number of parameters filled (so 1 in this case) on success.  It will return EOF if it can't match the pattern.

blogic:
merged into pending, Thanks !

PR #126, opened 2020-12-03
build: fix github runner incorrectly tags failed builds as passed

ArifAlam:
Fixes WIFI-979

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -64,5 +64,5 @@ cd -
 
 echo "### Building image ..."
 cd $BUILD_DIR
-make -j$(nproc) V=s 2>&1 | tee build.log
+make -j$(nproc) V=s
 echo "Done"


RickSommerville (commented):
Were we using the build.log file?  When I download the full raw logs from the build process, I don't see a file with this name but I do see huge files of name 1_build-ecw5211.txt (for instance).  These appear to possibly be target specific renames of build.log.

Please check that the raw build logs still contain the full output of the build process.

ArifAlam:
> Were we using the build.log file? When I download the full raw logs from the build process, I don't see a file with this name but I do see huge files of name 1_build-ecw5211.txt (for instance). These appear to possibly be target specific renames of build.log.
> 
> Please check that the raw build logs still contain the full output of the build process.

Confirmed that raw build logs still contain full output of the build process.  The raw build logs are generated by github runner.

blogic:
merged into pending, Thanks !

PR #131, opened 2020-12-08
Wifi-886 : Channels field in radio state table populated

AmmadRehmatCUS:
Wifi-886 : Channels field in radio state table populated

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/utils.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/utils.c
@@ -222,6 +222,58 @@ int phy_get_channels(const char *name, int *channel)
 	return j;
 }
 
+static void update_channels_state(struct schema_Wifi_Radio_State *rstate,
+			int *index, const char *key, int *value, int value_len)
+{
+	int y, z = 0;
+	char str[256];
+
+	STRSCPY(rstate->channels_keys[*index], key);
+
+	for (y = 0; y < value_len; y++)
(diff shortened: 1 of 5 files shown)


blogic:
the patch does not follow expected bracketing style and has whitespace errors.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/utils.c
```diff
@@ -222,6 +222,67 @@ int phy_get_channels(const char *name, int *channel)
...
+
+		if(l){
+			STRSCPY(rstate->channels_keys[ret], "allowed");
+			for (i=0; i<l; i++){
+   				index += sprintf(&str[index], "%d,", allow[i]);
+				}
+				memcpy(rstate->channels[ret], str, index);
+				index = 0;
+				ret++;
+		}
```
yashvardhan-netex, line 281: Just a thought. May be you can add a new function to populate the "channels" map. This would increase readability and you can get rid of some duplicate code. Something like this 

static void update_channels_state(struct schema_Wifi_Radio_State *rstate,
				      int *index, const char *key, const char *value)
{
	STRSCPY(rstate->channels_keys[*index], key);
	STRSCPY(rstate->channels[*index], value);

	*index = *index + 1;
	rstate->channels.len = *index;
}

And then simply call this above
if(j)
     update_channels_state(rstate, &index, "disabled", dis);
Similarily for other keys
Disclaimer: Code is not tested

AmmadRehmatCUS:
> the patch does not follow expected bracketing style and has whitespace errors.

fixed, please review and merge

PR #134, opened 2020-12-12
opensync: Capture dhcp packets in promiscous mode

yashvardhan-netex:
DHCP_leased_IP table was not getting populated for some of the
associated clients. Enabling capture in promiscous mode allows
capturing of all the dhcp packets.

Fixes: WIFI-1177


yashvardhan-netex:
This patch fixes a critical bug as DHCP_leased_table is used by cloud to get some of the associated clients info. Appreciate if this can be merged as early as possible.

greearb:
Maybe you should enable the #define instead of just removing it?  That is likely more maintainable going forward.

yashvardhan-netex:
> Maybe you should enable the #define instead of just removing it? That is likely more maintainable going forward.

Tried that before but somehow did not work as expected. So thought of taking up that later as an improvement. This fixes a crucial functionality so wanted this to go in.

yashvardhan-netex:
Wow! It was not working earlier because there was a bug in opensync. The config parameter was CONFIG_DHSNIFF_PCAP_PROMISC and the code was checking against CONFIG_INET_DHSNIFF_PCAP_PROMISC.
Fixed in the latest commit.

greearb:
John should be the one to review and this I think..

yashvardhan-netex:
> John should be the one to review and this I think..

Yeah. May be we can wait for him or if possible, can merge it(considering it addresses an important bug).  Either ways I will ask him to have a look when he is available. Thanks Ben!

blogic:
merged into pending, Thanks !

PR #135, opened 2020-12-13
Query DigiCert's API for Redirector's address

leonidlm:


feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
```diff
@@ -1,11 +1,30 @@
 #!/bin/sh
 
-if [ $# -ne 1 ] ; then
-	echo "Usage: $0 <redirector address>" >&2
-	exit 1
+AP_PRIVATE_KEY_FILE="/usr/opensync/certs/client_dec.key"
+AP_CERTIFICATE_FILE="/usr/opensync/certs/client.pem"
+AP_DEVICE_ID_FILE="/usr/opensync/certs/deviceid.txt"
```
RickSommerville, line 5: The file /usr/opensync/certs/deviceid.txt does not exist on my AP.  
What is the procedure to create this information.

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
```diff
@@ -1,11 +1,30 @@
...
+DIGICERT_API_URI="clientauth.demo.one.digicert.com"
```
RickSommerville, line 6: Is this URL correct?

RickSommerville (commented):
I am guessing there will be a new update to this PR before we are reading to actually apply on AP.

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
leonidlm: @brennerm can you please update the PR to reflect the correct Device ID file and also while you're here update the digicert API URL to production instance.

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
cbuschfb: TIP Root signed certificates must exist for all AP NOS / TIP NOS based devices. 

These function in mTLS therefore on each device 3 files are present:

device key

device certificate

CA certificate

This ensures a device is unique (key) is signed from same chain (device cert) and can validate the TLS handshake with server peer ( CA cert ). 

Note:

Device PKI files notes above MUST survive a factory reset 

Device MUST implement DigiCert device API lookup for current value of ‘redirector’ 
https://dev.digicert.com/digicert-one/iot-device-manager/iot-devices-v2/get-device-details-client-auth/ 

Which indicates the endpoint as: GET https://clientauth.one.digicert.com/iot/api/v2/device/{{device_id}}

brennerm:
Sorry folks, no idea why updating this branch closed this PR. Will create a new one.

PR #139, opened 2020-12-19
WIFI-1032 Repair EC420 MAC retrieval from manufacturing data

RickSommerville:


RickSommerville:
File is required because 11-ath10k-caldata uses it to correctly set the WiFI MAC addresses from the calibration data.

PR #140, opened 2020-12-20
firewall: Fix APC, UCC firewall rule addition

chaitanyagodavarthi:
Delete existing firewall rule before adding new
APC and UCC rule.


RickSommerville (commented):
Wouldn't it be better to make sure you only add once, instead of deleting and adding again?  This would prevent a change/reload on the firewall service.

I have seen 3 or more of each rule in some cases.  This delete will only remove one instance (I think) so it won't tidy up any currently messed up configs.

RickSommerville:
Thanks.  Looks good

RickSommerville (approved):
This version is better.

PR #141, opened 2020-12-21
Staging sartura dhcp events

lukakudra:
**NOTES:**
    ~~- this PR should be merged after https://github.com/Telecominfraproject/wlan-ap/pull/136.~~
    ~~- the code in this PR cannot be fully tested until the cloud is updated to support dhcp events protobuf definitions~~


lukakudra:
The code has been refactored based on the work done in [PR#147](https://github.com/Telecominfraproject/wlan-ap/pull/147) and [PR#148](https://github.com/Telecominfraproject/wlan-ap/pull/148).

yashvardhan-netex:
I gave it a run and the commit is crashing SM while processing DHCP packets. Can you look into it?

lukakudra:
> Seems dhcp_server_ip is not getting populated correctly.

The dhcp_server_ip field is populated only in some events like dhcp_ack_event or dhcp_offer_event, otherwise it's 0.0.0.0.
I think this has to do with the dhcp server itself because I see the same behaviour when i do a tcpdump.

lukakudra:
Thank you for the comments. I made the requested changes so you can take a look at the code again.

yashvardhan-netex:
> Thank you for the comments. I made the requested changes so you can take a look at the code again.

I will but has the crash being resolved? We must make the patch stable first.

lukakudra:
Thank you for the comments. I made the requested changes so you can take a look at the code again.

> > Thank you for the comments. I made the requested changes so you can take a look at the code again.
> 
> I will but has the crash being resolved? We must make the patch stable first.

The crashing is now resolved with the latest commit.
I decided to separate the `ubus_garbage_collector` function in `ubus_collector.c` into two separate functions: `ubus_session_garbage_collector` and `ubus_transaction_garbage_collector` that clear **client_sessions** and **dhcp_transactions** respectively.

lukakudra:
I added a limit to the number of dhcp events that could be cached in `dhcp_event_tree`.
The limit is hard-coded to 500, this is the same number that is used on hostapd for caching client connection events.

lukakudra:
> If the protobuf is being changed, the changes '**must**' be cascaded to the cloud/backend teams using the opensync_stats.proto at : https://github.com/Telecominfraproject/wlan-cloud-opensync-protobuf

Does that mean we need to open a PR with the changed protobuf at: https://github.com/Telecominfraproject/wlan-cloud-opensync-protobuf ?

mikehansen1970:
A pull request would be an excellent option as it would also prevent accidental clobber of parts of the proto that haven't changed. 
File is wlan-cloud-opensync-protobuf/src/main/protobuf/opensync_stats.proto
This file will need to be updated when the changes on the staging branch move into 'pending'. It is a bit of a balancing act, when things are changed, to get them in sync. For 'additions' it is less of a risk.

PR #147, opened 2020-12-24
opensync: Major refactoring/bug fixes for client events

yashvardhan-netex:
- Removed unwanted nested client event structures.
   Doing this not only reduced complexity in the code but also helped get
   rid of multple memory leaks and crashes in SM and datapipeline.
 - Simplified ubus invoke calls. Nested ubus invokes were resulting in some
 - data loss in callbacks
 - Fixed incorrect conversion of session_id data type. This was blocking
   clearing of sessions.
 - Got rid of unnecessary strdups/mallocs. Potential memory leaks.
 - Fixed missing conversion of frequency to band
 - Removed unwanted logging which was filling the syslog buffer.
 - Resolved a race occuring on the shared list between ubus_collector and SM.
 - Relevant changes in datapipeline/protobuf to support the modified interface.


yashvardhan-netex:
Sorry for the big patch. This sort of refactoring was required to get rid of the multiple issues seen in the client events code across different layers.

RickSommerville (approved):

PR #149, opened 2021-01-07
Force utilization >100% to 100%

RickSommerville:

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
@@ -266,7 +266,7 @@ bool target_stats_survey_get(radio_entry_t *radio_cfg, uint32_t *chan_list,
 	return ret;
 }
 
-#define PERCENT(v1, v2) (v2 > 0 ? (v1*100/v2) : 0)
+#define PERCENT(v1, v2) (v2 > 0 ? ((v1 > v2) ? 100 : (v1*100/v2)) : 0)
 
 bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t scan_type,
 				 target_survey_record_t *data_new, target_survey_record_t *data_old,


NagConnectus (approved):
very cool logic(:

greearb:
I'd think if it was significantly larger than 100% we'd want to know and try to debug root cause instead of just ignoring it....

RickSommerville:
Most of the time it looks like a couple of percent that I think may be due to sequential retrieval of the data.  For example:
"Tue Jan  5 19:08:49 2021 user.debug SM[2788]: <DEBUG>     MISC: Survey entry dur 41738 busy 42774 tx 42945 rx 50 chan 149 active 0 type 2".  The AP is very busy, basically transmitting 100% of the time.  The reported millisecond tick values increase slightly from duration < busy < tx.  This appears to be a repeating pattern  and is consistent with clear on read and individual operations to retrieve each count.

However, you are correct.  I am masking the problem not fixing it.  We could investigate why the raw data from nl80211 and the driver is inconsistent in more detail and try to fix it there.  I can raise a follow-up issue or can withdraw this pull-request.  Feed-back is welcome.

greearb:
I have noticed the percentages are off by a small amount sometimes, so as long as it is modestly small, then I think capping it to 100% sounds fine.  But if it is persistently off by a large amount, it would be good to know.

RickSommerville:
I traced the code and the data source is ultimately wmi.c/ath10k_wmi_event_pdev_bss_chan_info().  It gets all the data in one block from the firmware.  If that is not internally consistent, there is not much we can do except what we have done.

greearb:
There are small races and rounding errors in how the firmware gets info, and I'm not 100% confident that the register output is always perfect in how the firmware reads the raw info.  In my previous testing with ath10k channel utilization, it was generally close to what ath9k reported, and occasionally a few percentages over 100%.  My concern is that if there are ever regressions in the firmware/driver/whatever, your change might hide that.  And, for other chips that I am less confident in (ath11k, for instance), it may hide bugs.  That said, it is not a critical issue either way in my mind, so as long you feel it is good enough as is, then go ahead and push it and we can deal with problem cases in the future as needed.

RickSommerville:
Merged to pending.

PR #151, opened 2021-01-12
NETEXP-928: fix associated clients keeps switching state

ArifAlam:
Fixes an issue where client was getting deleted and added to ovsdb causing state change.


RickSommerville (commented):
Do the LOG calls at lines 281 and 298 need to be LOGI?  I suggest we drop them to LOGD.

We don't need the blank lines at 284 and 301.

RickSommerville:
Thanks, merged fix.

PR #152, opened 2021-01-13
wlan-ap-consumer

lkristan-plume:
Access point implementation using cfg80211 platform and standard OpenSync services.


blogic:
Hi
kindly consider to rework the patch annotation of all 3 patches
"wlan-ap-consumer: $subject

$description

Signed-off-by: Name <email>"

Thanks !

blogic:
@lkristan-plume Hi
static code analyses spotted various errors, including null pointer derefs
https://github.com/Telecominfraproject/wlan-ap/pull/152/checks?check_run_id=1927307209

lkristan-plume:
> static code analyses spotted various errors, including null pointer derefs
@blogic, AFAICS the failures are in parts of code not touched by our commits. I presume the static analysis checks were introduced recently, and the offending parts of code were not fixed yet on trunk.

matej-plume:
@blogic any update on static resolution ... we would like to get this patch approved asap since we have Adtran waithing with BCM and MTK contributions

lkristan-plume:
@blogic, PR has been rebased and updated

cbuschfb (commented):
Hi @mrajsek-plume  in these commits OVS is re-added into the system. The SDK would not know how to handle this behaviour the resulting VIF : linux bridge state currently known in TIP would possibly become lost. We do not have any CI sanity testing for this. Could you clarify further the OVS bridge additions as they realte to behaviour in the system?

matej-plume:
Hi @cbuschfb Please note that 'consumer' product and its services are implemented as a separate feed and do not modify any of the code used for 'enterprise' mainstream products. The main change regarding the use of OVS is in this commit: https://github.com/Telecominfraproject/wlan-ap/pull/152/commits/0ca28d37a1a999a9bffb934e8586c697f479f062. The change is not affecting API therefore it should be agnostic also to the cloud, but at this point in time several of 'consumer' advanced services rely on OVS and use standard OpenSync managers OM, FSM, FCM which was the reason for adding it on-top. 

At the moment 'consumer' and 'enterprise' products (feeds) differ only on some "service" needed code where majority is the same and as we agreed we shall work towards merging common and OpenSync code as we should also work towards adding CI sanity testing.

PR #153, opened 2021-01-14
opensync: Fix events timestamp

yashvardhan-netex:
- Added milliseconds timestamp precision
- Replaced realtime clock with monotonic
- Data type changed to uint64 to hold millisecond timestamp

### DIFF
--- a/feeds/wifi-trunk/hostapd/patches/900-hapd-netlink-ubus-bridge.patch
+++ b/feeds/wifi-trunk/hostapd/patches/900-hapd-netlink-ubus-bridge.patch
@@ -184,7 +184,7 @@
  
  static inline struct hapd_interfaces *get_hapd_interfaces_from_object(struct ubus_object *obj)
  {
-@@ -63,6 +66,16 @@ static void hostapd_ubus_connection_lost
+@@ -63,6 +66,21 @@ static void hostapd_ubus_connection_lost
  	eloop_register_timeout(1, 0, ubus_reconnect_timeout, ctx, NULL);
  }
  
@@ -198,10 +198,15 @@
 +		return *id1 > *id2;
 +}
 +
++uint64_t get_time_in_ms(struct timespec *ts)
++{
++    return (uint64_t) ts->tv_sec * 1000 + ts->tv_nsec / 1000000;
++}
++
  static bool hostapd_ubus_init(void)
  {
  	if (ctx)
-@@ -525,6 +538,177 @@ static const struct blobmsg_policy csa_p
+@@ -525,6 +543,177 @@ static const struct blobmsg_policy csa_p
  };
  
  #ifdef NEED_AP_MLME
@@ -297,7 +302,7 @@
 +				struct client_auth_event *p = &c_rec->u.auth;
 +				t_msg = blobmsg_open_table(&b_ev, "ClientAuthEvent");
 +				blobmsg_add_u64(&b_ev, "session_id", rec->session_id);
-+				blobmsg_add_u32(&b_ev, "timestamp", c_rec->ts.tv_sec);
++				blobmsg_add_u64(&b_ev, "timestamp", c_rec->timestamp);
 +				blobmsg_add_string(&b_ev, "sta_mac", p->sta_mac);
 +				blobmsg_add_u32(&b_ev, "band", p->band);
 +				blobmsg_add_u32(&b_ev, "auth_status", p->auth_status);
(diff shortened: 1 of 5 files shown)


feeds/wifi-trunk/hostapd/patches/900-hapd-netlink-ubus-bridge.patch
```diff
@@ -488,7 +493,7 @@
 +	}
 +	struct hostapd_event_avl_rec *rec = NULL;
 +	struct timespec ts;
-+	clock_gettime(CLOCK_REALTIME, &ts);
++	clock_gettime(CLOCK_MONOTONIC, &ts);
```
RickSommerville, line 496: Does this change align with Cloud side expectations?

feeds/wlan-ap/opensync/patches/19-events-protobuf-definitions.patch
```diff
@@ -106,7 +104,7 @@ Index: opensync-2.0.5.0/interfaces/opensync_stats.proto
 +        optional bool                   using11k                            = 9;
 +        optional bool                   using11r                            = 10;
 +        optional bool                   using11v                            = 11;
-+        optional uint32                 timestamp_ms                        = 12;
++        required uint64                 timestamp_ms                        = 12;
```
RickSommerville, line 107: I recall the suggestion that the timestamp_ms should be a required field came from the Cloud side.  Is the change coordinated with them?

feeds/wlan-ap/opensync/patches/19-events-protobuf-definitions.patch
```diff
@@ -1,7 +1,5 @@
-Index: opensync-2.0.5.0/interfaces/opensync_stats.proto
-===================================================================
---- opensync-2.0.5.0.orig/interfaces/opensync_stats.proto
-+++ opensync-2.0.5.0/interfaces/opensync_stats.proto
+--- a/interfaces/opensync_stats.proto
```
RickSommerville, line 1: Is the change to the path here significant?  Applies to both patches.

feeds/wlan-ap/opensync/patches/20-dppline-event.patch
```diff
@@ -1,7 +1,5 @@
-Index: opensync-2.0.5.0/src/lib/datapipeline/inc/dppline.h
-===================================================================
---- opensync-2.0.5.0.orig/src/lib/datapipeline/inc/dppline.h
-+++ opensync-2.0.5.0/src/lib/datapipeline/inc/dppline.h
+--- a/src/lib/datapipeline/inc/dppline.h
++++ b/src/lib/datapipeline/inc/dppline.h
 @@ -47,6 +47,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBI
```
RickSommerville, line 3: Path change again.

feeds/wifi-trunk/hostapd/patches/900-hapd-netlink-ubus-bridge.patch
yashvardhan-netex: I don't think cloud would have any issues with this change. The wall clock(CLOCK_REALTIME) can jump forwards and backwards when the local clock is changed for example by NTP. MONOTONIC, on the other hand will make sure time always moves forward and thus is the best option when we are trying to calculate elapsed time between two events.

feeds/wlan-ap/opensync/patches/19-events-protobuf-definitions.patch
yashvardhan-netex: I think this is because I have changed the existing patch. The regular quilt process. Doesn't seem to be an issue.

feeds/wlan-ap/opensync/patches/20-dppline-event.patch
yashvardhan-netex: Same here

feeds/wifi-trunk/hostapd/patches/900-hapd-netlink-ubus-bridge.patch
yashvardhan-netex: I also looked at the code in the reference product and there also we are using monotonic clock.

yashvardhan-netex: Since cloud is expecting events to be reported with clock time so that it can be aligned across systems, I have pushed a new commit reverting back to realtime

RickSommerville (approved):

RickSommerville:
Merged, thanks.

PR #160, opened 2021-01-28
build: generate sysupgrade.bin instead of sysupgrade.tar for AX images

ArifAlam:
Generate sysupgrade.bin to be consistant with sysupgrade image naming.

### DIFF
--- a/patches/0027-ipq807x-add-the-Qualcomm-AX-target-support.patch
+++ b/patches/0027-ipq807x-add-the-Qualcomm-AX-target-support.patch
@@ -7324,8 +7324,8 @@ index 0000000000..3a219e6d6c
 +  ROOTFSNAME_IN_UBI := ubi_rootfs
 +  BLOCKSIZE := 128k
 +  PAGESIZE := 2048
-+  IMAGES := sysupgrade.tar nand-factory.bin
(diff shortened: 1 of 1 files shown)


patches/0027-ipq807x-add-the-Qualcomm-AX-target-support.patch
```diff
@@ -7324,8 +7324,8 @@ index 0000000000..3a219e6d6c
 +  ROOTFSNAME_IN_UBI := ubi_rootfs
 +  BLOCKSIZE := 128k
 +  PAGESIZE := 2048
-+  IMAGES := sysupgrade.tar nand-factory.bin
-+  IMAGE/sysupgrade.tar := sysupgrade-tar | append-metadata
++  IMAGES := sysupgrade.bin nand-factory.bin
++  IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
```
RickSommerville, line 7328: Is this just renaming the file?
What will happen when sysupgrade is called and the content is in a different format?

blogic:
why is this change needed ?

ArifAlam:
Taking a different approach.  Closing this PR.

PR #167, opened 2021-02-03
Staging command manager deletions

AmmadRehmatCUS:
WIFI-1412 Command manager clears entries upon completion/failure from state and config table
    

AmmadRehmatCUS:
not required, commit added to stop session branch as history was not correct on this one

PR #171, opened 2021-02-08
WIFI-1458-Tx-Power-Range

NagConnectus:
This patch will resolve the radio config fail problem, when tx
power configured more than allowed operational range of AP.


RickSommerville (approved):

RickSommerville:
Merged, thanks

PR #177, opened 2021-02-17
opensync: Update ovsdb schema to support 11ax

yashvardhan-netex:
Ovsdb schema for VIF_Config/State is updated to support 11ax as min_hw_mode

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
@@ -119,6 +119,7 @@ enum {
 	WIF_ATTR_DVLAN_FILE,
 	WIF_ATTR_DVLAN_NAMING,
 	WIF_ATTR_DVLAN_BRIDGE,
+	WIF_ATTR_MIN_HW_MODE,
 	__WIF_ATTR_MAX,
 };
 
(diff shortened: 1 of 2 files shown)


yashvardhan-netex:
This still needs to be tested against Opensync Gateway. I will ask Mike to test the staging load before it can be merged.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -758,12 +759,8 @@ bool vif_state_update(struct uci_section *s, struct schema_Wifi_VIF_Config *vcon
...
 		LOGN("%s: Failed to get channel", vstate.if_name);
 
-	phy_get_band(target_map_ifname(radio), band);
-	LOGD("Find min_hw_mode: Radio: %s Phy: %s Band: %s", radio, target_map_ifname(radio), band );
-	if (strstr(band, "5"))
-		SCHEMA_SET_STR(vstate.min_hw_mode, "11ac");
-	else
-		SCHEMA_SET_STR(vstate.min_hw_mode, "11n");
+	if (tb[WIF_ATTR_MIN_HW_MODE])
+		SCHEMA_SET_STR(vstate.min_hw_mode, blobmsg_get_string(tb[WIF_ATTR_MIN_HW_MODE]));
```
RickSommerville, line 763: I don't think UCI actually uses this for anything so why write it there.
If necessary, we can stash it in the phy_tree defined in radio_nl80211.c.
I don't know what Mike is doing with this but we are initializing it from the initial update of Wifi_Radio_State and Wifi_Radio_Config on start-up.  I don't think we can retrieve it from UCI because no-one will put it there in the first place.

We can probably initialize it on start-up based on data found by radio_nl80211.c/nl80211_add_phy().  Would need to look at capabilities fields.

RickSommerville (changes requested):

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
yashvardhan-netex: The idea here is to satisfy the opensync interface and make sure there is no mismatch between Radio and VIF Config for this parameter. Other than that I do not think there is no much use of this parameter in VIF Config as it is a radio parameter. The parameter would be set inside the VIF config by the Gateway based on the radio config. I had the load tested with Mike and he has given a go ahead.

RickSommerville: Thanks for explanation and test results.

RickSommerville (approved):

RickSommerville:
Merged, thanks

PR #187, opened 2021-02-22
Staging wifi 1595 dfs channel list

AmmadRehmatCUS:
Staging wifi 1595 dfs channel list

-  channels list only has backup_channel in order to control which channel should the hostapd choose in case of DFS event.
-  5G frequencies are treated as only 5G radios and not 5GU/5GL

### DIFF
--- a/feeds/wifi-trunk/hostapd/patches/901-hapd-ubus-channel-switch.patch
+++ b/feeds/wifi-trunk/hostapd/patches/901-hapd-ubus-channel-switch.patch
@@ -140,8 +140,8 @@ Index: hostapd-2020-06-08-5a8b3662/src/ap/ubus.h
  };
  
 +enum hostapd_ubus_chan_event_reason {
-+	HOSTAPD_UBUS_HIGH_INTERFERENCE,
-+	HOSTAPD_UBUS_DFS_SWITCH
++	HOSTAPD_UBUS_DFS_SWITCH,
++	HOSTAPD_UBUS_HIGH_INTERFERENCE
 +};
 +
  struct hostapd_ubus_request {
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -274,7 +274,6 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
 
 	char phy[6];
 	char ifname[8];
-	int list_channels[IEEE80211_CHAN_MAX] , list_channels_len = 0;
 
 	strncpy(ifname, rconf->if_name, sizeof(ifname));
 	strncpy(phy, target_map_ifname(ifname), sizeof(phy));
@@ -342,14 +341,12 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
 			 LOGE("%s: failed to set ht/hwmode", rconf->if_name);
 	}
 
-	list_channels_len = phy_get_list_channels_dfs(phy, list_channels);
-	if (list_channels_len) {
-		struct blob_attr *n;
+	struct blob_attr *n;
+	int backup_channel = 0;
+	backup_channel = rrm_get_backup_channel(rconf->freq_band);
+	if(backup_channel) {
 		n = blobmsg_open_array(&b, "channels");
-		for(int i = 0; i < list_channels_len; i++)
-		{
-			blobmsg_add_u32(&b, NULL, list_channels[i]);
-		}
+		blobmsg_add_u32(&b, NULL, backup_channel);
 		blobmsg_close_array(&b, n);
 	}
 
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
@@ -53,6 +53,22 @@ void rrm_config_vif(struct blob_buf *b, struct blob_buf *del, const char * freq_
 	return;
 }
 
+int rrm_get_backup_channel(const char * freq_band)
+{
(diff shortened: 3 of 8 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -342,16 +341,11 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
...
-		for(int i = 0; i < list_channels_len; i++)
-		{
-			blobmsg_add_u32(&b, NULL, list_channels[i]);
-		}
-		blobmsg_close_array(&b, n);
-	}
+	struct blob_attr *n;
+	n = blobmsg_open_array(&b, "channels");
+	blobmsg_add_u32(&b, NULL, rconf->channel);
+	blobmsg_add_u32(&b, NULL, rrm_get_backup_channel(rconf->freq_band));
```
RickSommerville, line 346: Does it matter that this is now an option instead of a list with 1 entry?

The call to rrm_get_backup_channel() can fail returning 0.  What happens in hostapd when we set this channel to 0?  Previously, we had 'channels' set the same as 'channel' due to scripting.  Would that be a safer option in this case.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
```diff
@@ -53,6 +53,22 @@ void rrm_config_vif(struct blob_buf *b, struct blob_buf *del, const char * freq_
...
+int rrm_get_backup_channel(const char * freq_band)
+{
+	struct schema_Wifi_RRM_Config conf;
+
+	memset(&conf, 0, sizeof(conf));
+	if (false == ovsdb_table_select_one(&table_Wifi_RRM_Config,
+			SCHEMA_COLUMN(Wifi_RRM_Config, freq_band), freq_band, &conf))
+	{
+		LOG(ERR, "Wifi_Radio_Config: No RRM for band %s", freq_band );
+		return 0;
```
RickSommerville, line 65: This is the failure I mentioned above.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
```diff
@@ -96,12 +112,35 @@ static bool rrm_config_changed( struct schema_Wifi_RRM_Config *old,
...
+
+	return true;
+}
+
 static bool rrm_config_set( struct schema_Wifi_RRM_Config *old,
 	struct schema_Wifi_RRM_Config *conf ) 
 {
 	if (rrm_config_changed(old, conf)) {
-		return( rrm_config_update(conf, true));
+		rrm_config_update(conf, true);
```
RickSommerville, line 137: Lost the ability to report an error upstream.  Could create 'bool rc = true;' and "AND" it with the return from the two functions. 

I don't think it matters because we discard the return code a layer up.

RickSommerville (commented):
See comments

feeds/wifi-trunk/hostapd/patches/901-hapd-ubus-channel-switch.patch
```diff
@@ -140,8 +140,8 @@ Index: hostapd-2020-06-08-5a8b3662/src/ap/ubus.h
  };
  
 +enum hostapd_ubus_chan_event_reason {
-+	HOSTAPD_UBUS_HIGH_INTERFERENCE,
-+	HOSTAPD_UBUS_DFS_SWITCH
++	HOSTAPD_UBUS_DFS_SWITCH,
```
RickSommerville, line 143: What does this change do?

RickSommerville:
Also these two commits should be squashed.  Second is dependant on first and first is incorrect without the second.

Both defects can refer to same commit for closure.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
AmmadRehmatCUS: this is to avoid returning if both rrm_config_changed and rrm_radio_config_update conditions are true , we dont want to return from here without executing the rest of the function

feeds/wifi-trunk/hostapd/patches/901-hapd-ubus-channel-switch.patch
AmmadRehmatCUS: this is to align with how the enum is defined in the protobufs

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
AmmadRehmatCUS: @RickSommerville  I can add a check before writing the backup channel that its not 0, by providing the current channel alone in the list we would limit the AP from switching in case of radar event,

feeds/wlan-ap/opensync/src/src/sm/src/ubus_collector.c
```diff
@@ -126,13 +126,13 @@ static int frequency_to_channel(int freq)
 static radio_type_t frequency_to_band(int freq)
 {
 	int chan = frequency_to_channel(freq);
-
+	/*Need to add support for differentiating between 5G,5GU and 5GL*/
```
RickSommerville, line 129: This function is also used for the Client events: authenticate, associate, and disconnect.
Did you intend to change their behaviour?
Did they have the same problem?
Did you test that this better for them?

AmmadRehmatCUS: I assumed that since we are supporting 2 radio device only for release 1 it would be same for client events, although some regression testing will be required. If my assumption is incorrect we will have to make 2 separate functions although in case of client events we would face the same problem on 2 radio devices that radios will be reported as 5GU/5GL no 5G.

PR #193, opened 2021-02-24
Wifi-1623 : Any event coming from hostapd with 5G frequency will be t…

AmmadRehmatCUS:
Wifi-1623 : Any event coming from hostapd with 5G frequency will be treated as 5G radio only

-  channels list only has backup_channel in order to control which channel should the hostapd choose in case of DFS event.
-  5G frequencies are treated as only 5G radios and not 5GU/5GL

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -274,7 +274,6 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
 
 	char phy[6];
 	char ifname[8];
-	int list_channels[IEEE80211_CHAN_MAX] , list_channels_len = 0;
 
 	strncpy(ifname, rconf->if_name, sizeof(ifname));
 	strncpy(phy, target_map_ifname(ifname), sizeof(phy));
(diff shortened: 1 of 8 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -342,14 +341,12 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
...
+	backup_channel = rrm_get_backup_channel(rconf->freq_band);
+	if(backup_channel) {
 		n = blobmsg_open_array(&b, "channels");
-		for(int i = 0; i < list_channels_len; i++)
-		{
-			blobmsg_add_u32(&b, NULL, list_channels[i]);
-		}
+		blobmsg_add_u32(&b, NULL, backup_channel);
 		blobmsg_close_array(&b, n);
 	}
```
RickSommerville, line 351: Should we be deleting the list if no backup channel is supplied?  Couldn't this result in stale data?  Deleting the list would leave the old default channels = channel behaviour.

RickSommerville: Implemented.

RickSommerville:
Merged, thanks

PR #195, opened 2021-02-25
Wifi-1572 Repair vif state to show the correct number of associated clients

yashvardhan-netex:
This patch primarily reverts fdf5bb6e96f0b232a99903777f628769ec47aa55
(opensync: add average STA RSSI tracking). Netlink request to track STA RSSI
every 10 secs was leading to a race condition because the receive callback
for the request was also used to receive rest of the nl80211 events.
This was resulting in lost messages and hence VIF State was not getting populated correctly.
Now, the code to poll for STA RSSI in opensync is redundant and is not used because STA RSSI is already
tracked in hostapd. Removing it solves the redundancy and the race.


RickSommerville (approved):
Looks good.

RickSommerville:
Merged, thank you.

PR #198, opened 2021-02-25
Wifi-1470 Fix missing ip string to bytes conversion

yashvardhan-netex:
IP event protobuf expects ip address to be in the form of bytes.
Added the missing conversion.


RickSommerville (approved):

RickSommerville:
Merged, thanks

PR #202, opened 2021-03-04
wifi-1614 Disable-TX_STBC-when-configuring-1x1

netexp-ravi:
Added support to update the TX-STBC capability if only single Tx-antenna is available.

### DIFF
--- a/feeds/wifi-ax/mac80211/files/lib/netifd/wireless/mac80211.sh
+++ b/feeds/wifi-ax/mac80211/files/lib/netifd/wireless/mac80211.sh
@@ -254,6 +254,11 @@ mac80211_hostapd_setup_base() {
 		[ "$rx_stbc" -lt "$cap_rx_stbc" ] && cap_rx_stbc="$rx_stbc"
 		ht_cap_mask="$(( ($ht_cap_mask & ~(0x300)) | ($cap_rx_stbc << 8) ))"
 
+		conf_tx_ant=$(iw phy "$phy" info | grep 'Configured Antennas:' | cut -d: -f2)
+		set -- $conf_tx_ant
+		if [ "$2" == "0x1" -o "$2" == "0x10" ]; then
+			tx_stbc=0
+		fi
 		mac80211_add_capabilities ht_capab_flags $ht_cap_mask \
 			LDPC:0x1::$ldpc \
 			GF:0x10::$greenfield \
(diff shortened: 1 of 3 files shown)


feeds/wifi-ax/mac80211/files/lib/netifd/wireless/mac80211.sh
```diff
@@ -343,6 +348,11 @@ mac80211_hostapd_setup_base() {
 		[ "$rx_stbc" -lt "$cap_rx_stbc" ] && cap_rx_stbc="$rx_stbc"
 		vht_cap="$(( ($vht_cap & ~(0x700)) | ($cap_rx_stbc << 8) ))"
 
+		conf_tx_ant=$(iw phy "$phy" info | grep 'Configured Antennas:' | cut -d: -f2)
+		set -- $conf_tx_ant
+		if [ "$2" == "0x1" ]; then
+			tx_stbc_2by1=0
```
RickSommerville, line 354: Some of the Wifi6 APs seem to use the upper nibble of Configured Antennas instead of the lower nibble.  Example is EX447 (Enis has 1 you can test on).  As a result the test may need to be (addional bracket for order of operation?)
if [ "$2" == "0x1" -o "$2" == "0x10" ];

RickSommerville (commented):
Need to test on an EX447.  See other comment.

Also, in the defect please put the Root Cause, Solution and Tests in the Description not in the Comment.

feeds/wifi-ax/mac80211/files/lib/netifd/wireless/mac80211.sh
netexp-ravi: Done.

RickSommerville (approved):

RickSommerville:
Merged, thank you

blogic:
could you explain the problem here to me ? this should be upstreamed if there is an issue with the mac80211.sh script

RickSommerville:
STBC space time block coding requires multiple antennas.  When the MIMO mode is configured as 1x1, or more specifically when only 1 TX antenna is enabled, then TX-STBC must be disabled.

blogic:
I looked into this. the patch is a workaround for the real bug.
the driver should update its caps when going into 1x1
which driver do you see this issue on ?

RickSommerville:
Ath10k and ath11k

On Mar. 13, 2021 3:34 a.m., John Crispin ***@***.***> wrote:

I looked into this. the patch is a workaround for the real bug.
the driver should update its caps when going into 1x1
which driver do you see this issue on ?

PR #203, opened 2021-03-05
trigger release build on Git tag push

brennerm:


.github/workflows/build.yml
```diff
@@ -19,6 +21,8 @@ jobs:
           GH_BUILD_PASSWORD: ${{ secrets.GH_BUILD_PASSWORD }}
       run: |
         TAG=$(date +%Y-%m-%d)-$(git rev-parse --short HEAD)
+        [[ "${{ github.ref }}" == "refs/tags/"* ]] && TAG=$(echo "${{ github.ref }}" | sed 's/refs\/tags\/[v]//')
```
ArifAlam, line 24: Just want to confirm lines similar to line 24 will only get run on tagged commits and that untagged commits will use the TAG from line 23.

brennerm: Yup, that's the intention. :+1:

blogic:
the expected PR will contain within its patches the following format
$prefix: $descprition
$explanation
Signed-off-by: $name <$email>

brennerm:
@blogic Done :+1:

RickSommerville:
Merged, thanks

PR #209, opened 2021-03-12
WIFI-1629 Captive Portal Service Halt

NagConnectus:
This patch will add the support to enable captive portal on multiple ssid's with same
captive portal profile

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/captive.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/captive.c
@@ -158,7 +158,7 @@ void vif_state_dhcp_allowlist_get(struct schema_Wifi_VIF_State *vstate)
 			if (blobmsg_type(cur) != BLOBMSG_TYPE_STRING)
 				continue;
 			sscanf(blobmsg_get_string(cur), "/%[^/]/%[^_]_%s", fqdn, set, read_ifname);
-			if(!strcmp(vstate->if_name, read_ifname))
+			if(!strcmp("opennds", read_ifname))
 			{
 				strcpy(vstate->captive_allowlist[vstate->captive_allowlist_len], fqdn);
 				vstate->captive_allowlist_len++;
@@ -171,10 +171,8 @@ void vif_state_dhcp_allowlist_get(struct schema_Wifi_VIF_State *vstate)
 
 void ipset_flush(char *ifname)
 {
-	char com[24]="ipset flush set_wlan1";
-
-	if (!strcmp(ifname,"wlan1"))
-		system(com);
+	char com[32]="ipset flush set_opennds";
+	system(com);
 }
 
 void vif_dhcp_opennds_allowlist_set(const struct schema_Wifi_VIF_Config *vconf, char *ifname)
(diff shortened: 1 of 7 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -803,9 +806,10 @@ bool vif_state_update(struct uci_section *s, struct schema_Wifi_VIF_Config *vcon
 	}
 	vif_state_security_get(&vstate, tb);
 	vif_state_custom_options_get(&vstate, tb);
-	vif_state_captive_portal_options_get(&vstate, s);
-	vif_state_dhcp_allowlist_get(&vstate);
-
+	if(!strcmp(network_name,"lan"))
+		vif_state_captive_portal_options_get(&vstate);
+	if(!strcmp(network_name,"lan"))
```
RickSommerville, line 811: Why did the dhcp-allowlist get pushed into this "if" statement at line 811?  If it needs to be in the "if", then the "if" at 809 is identical and. should be used for both.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/captive.c
```diff
@@ -228,134 +226,133 @@ void set_captive_portal_state(struct schema_Wifi_VIF_State *vstate,
...
 	int i;
 	int index = 0;
 	const char *opt;
 	char *buf = NULL;
 	struct blob_attr *tc[__NDS_ATTR_MAX] = { };
-	struct uci_element *e = NULL;
+	struct uci_section *cp_section;
 
 	uci_load(cap_uci, "opennds", &opennds);
-	uci_foreach_element(&opennds->sections, e) {
```
RickSommerville, line 241: This large change is really mostly white-space.  It would be much easier to review if you had replaced the "if !strcmp()) {" at line 243 with just a "{".  Then the indentation could stay the same.

I don't actually suggest you do this now.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
NagConnectus: we have captive_allowlist and captive_portal, two separated entires in the vif config and vif state table. When that respective entry changed, then only I'm trying to check the condition and asking to update the state

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/captive.c
NagConnectus: got it, thank you

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
RickSommerville: So you are saying it is correct for them both to depend on "network_name==lan".  Correct?
If the condition is the same, then don't test it twice;  line 810 and 812 should be in the same if statement at line 809.

NagConnectus: yes, I'll quickly change and update

RickSommerville:
Merged, thank you

PR #210, opened 2021-03-15
openwrt: Nand images for tp-link EX447 and EX227

chaitanyagodavarthi:
Generate factory nand images for tp-link ex447 and ex227.
This image is used to flash for migrating from stock qsdk
image to tip openwrt image using the nand flash.

### DIFF
--- a/patches/0045-tp-link-ex447-ex227-factory-Nand-images.patch
+++ b/patches/0045-tp-link-ex447-ex227-factory-Nand-images.patch
@@ -1,0 +1,70 @@
+From 57a733a35f7727d2e1b786e4270de7d593d90614 Mon Sep 17 00:00:00 2001
+From: Chaitanya Godavarthi <chaitanya.kiran@netexperience.com>
+Date: Fri, 12 Mar 2021 15:39:20 -0500
+Subject: [PATCH] openwrt: Generate factory Nand images for tp-link EX447 and
+ EX227
+
+Generate factory nand images for tp-link ex447 and ex227.
+This image is used to flash for migrating from stock qsdk
+image to tip openwrt image using the nand flash.
+---
+ target/linux/ipq807x/base-files/lib/upgrade/platform.sh | 8 ++++----
+ target/linux/ipq807x/image/ipq807x.mk                   | 6 ++++--
+ 2 files changed, 8 insertions(+), 6 deletions(-)
+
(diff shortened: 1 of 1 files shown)


patches/0045-tp-link-ex447-ex227-factory-Nand-images.patch
```diff
@@ -0,0 +1,70 @@
...
+diff --git a/target/linux/ipq807x/image/ipq807x.mk b/target/linux/ipq807x/image/ipq807x.mk
+index 8f9cda14a3..a41097e08e 100644
+--- a/target/linux/ipq807x/image/ipq807x.mk
++++ b/target/linux/ipq807x/image/ipq807x.mk
+@@ -50,9 +50,10 @@ define Device/tplink_ex227
+   DEVICE_DTS_CONFIG=config@hk07
+   SUPPORTED_DEVICES := tplink,ex227
+   DEVICE_PACKAGES := ath11k-wifi-tplink-ex227
+-  IMAGES := sysupgrade.tar nor-factory.bin
++  IMAGES := sysupgrade.tar nor-factory.bin nand-factory.bin
```
RickSommerville, line 50: Is nor-factory still supposed to be here?  Multiple places.

RickSommerville (commented):
Is upgrade procedure documented in Confluence?  Please reply with the page link.

patches/0045-tp-link-ex447-ex227-factory-Nand-images.patch
chaitanyagodavarthi: If we need to generate the nor based images along with the Nand based images then this is needed.

chaitanyagodavarthi:
> Is upgrade procedure documented in Confluence? Please reply with the page link.

https://telecominfraproject.atlassian.net/wiki/spaces/WIFI/pages/947978599/EX227+and+EX447

patches/0045-tp-link-ex447-ex227-factory-Nand-images.patch
RickSommerville: Only 1 sysupgrade.tar is generated and it is the same for both NOR and NAND upgrade?  Is this correct?

RickSommerville:
Merged 2cc526b, thanks

PR #217, opened 2021-03-22
wifi-1737: Restrict DHCP Sniffing to a max of 4 vlans

yashvardhan-netex:
Partial fix for wifi-1737:
Use of current DHCP sniffing library from opensync increases
memory footprint of nm with increased number of vlans.
This patch restricts DHCP sniffing on a maximum of 4 vlans.
This is a workaround and actual fix would be to use a better
packet filter such as eBPF

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/netifd/src/inet_conf.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/netifd/src/inet_conf.c
@@ -6,30 +6,43 @@
 
 #include "inet_conf.h"
 
-struct netifd_iface* netifd_add_inet_conf(struct schema_Wifi_Inet_Config *iconf)
-{
+static int vlan_count = 0;
 
+void netifd_add_inet_conf(struct schema_Wifi_Inet_Config *iconf)
+{
 	struct netifd_iface *piface = NULL;
 
-	piface = netifd_iface_get_by_name(iconf->if_name);
-	if (piface == NULL)
-	{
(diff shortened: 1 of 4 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/netifd/src/inet_conf.c
```diff
@@ -6,30 +6,38 @@
...
-		{
-			LOGN("Setting up dhsnif for %s", piface->if_base->inet.in_ifname);
+		if (!strcmp(iconf->if_name, "wan") || !strcmp(iconf->if_name, "lan")) {
 			netifd_inet_config_set(piface);
 			netifd_inet_config_apply(piface);
+		} else if (iconf->vlan_id_exists && iconf->vlan_id > 2) {
+			if (vlan_count < DHCP_SNIFF_MAX_VLAN && !strstr(iconf->if_name,"lan_")) {
+				netifd_inet_config_set(piface);
+				netifd_inet_config_apply(piface);
+				vlan_count++;
```
RickSommerville, line 36: I don't see anything that decrements this count.  If you delete you config and add a new one, does the DHCP sniff get started?

yashvardhan-netex: Fixed with the new patch. Tested deletion and addition of a new config, Sniffing is always enabled on a max of 4 vlans.

RickSommerville (approved):
Looks good

RickSommerville:
Merged 9d7beae, thanks

PR #219, opened 2021-03-23
Staging wifi 464 handling manufacturing data

netexp-ravi:
Adding new manufacturing data information to the AWLAN Node.
Still need to take care of "qr_code". Tested on Edgecore ECW5211.


RickSommerville:
Should be a pull against pending not trunk.

PR #224, opened 2021-03-30
openwrt: Revert UBUS pending list iteration method

chaitanyagodavarthi:
This commit partially reverts the earlier patch:
"workaround possibly false positive uses of memory after it is freed"

This fixed the crash while freeing a list item in
ubus_process_pending_msg due to a double free.


blogic:
merged in pending, Thanks !

PR #227, opened 2021-04-01
Wifi-1720. Fix for the neighbor scan failure on wifi6 APs.

netexp-ravi:
Change contain John Crispin's patch to copy the nl80211-ax.h and use it for wifi6 APs.

Tested on WF188n (wifi6) and ECW5211 (wifi5) and EA8300 (wifi5) APs.
Checked for the scan results in EquipmentScanDetails event on the cloud.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
@@ -547,7 +547,7 @@ int nl80211_scan_trigger(struct nl_call_param *nl_call_param, uint32_t *chan_lis
 {
 	struct nl_msg *msg = nl80211_call_vif(nl_call_param, NL80211_CMD_TRIGGER_SCAN, false);
 	struct nlattr *freq;
-	unsigned int i;
+	unsigned int i, flags = 0;
 	int ret = 0;
 
 	if (!msg)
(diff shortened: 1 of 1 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio_nl80211.c
```diff
@@ -14,7 +14,7 @@
 #include <sys/types.h>
 
 #include <linux/sockios.h>
-#include <linux/nl80211.h>
+#include "nl80211-copy.h"
```
yashvardhan-netex, line 17: If we are using same header for WIFI-5 and WIFI-6, what is the point of checking in two headers?

yashvardhan-netex (commented):
Did we also test client Events?

netexp-ravi:
> Did we also test client Events?

Yes, we tested the client events on wifi5 (ECW5211) AP.

yashvardhan-netex (approved):
Looks good.

blogic:
Does scan work with this patch ?

netexp-ravi:
> Does scan work with this patch ?

Yes, scan works with this patch. We checked the NEIGHBOR_SCAN events on the cloud with other AP information in the event. I am still testing for other stuff like client events.

RickSommerville (approved):
Looks good.  Glad we didn't require the alternate nl80211.h.

RickSommerville:
Merged 077e1cd, thanks.

PR #230, opened 2021-04-08
Wifi-1914. Fix for FT-EAP roaming failure.

netexp-ravi:
Adding the required uci options for FT-EAP roaming.

- adding r0kh and r1kh uci options which gets propagated to hostapd.
- updating nasid with BSSID if nothing is configured from cloud. This does leave us with mismatched vifS and vifC.

With these changes, we do not see the Authentication failure anymore for mismatched PMKID. 
But we see some Roam-test failures due to Authentication timeout. Need to investigate more in that.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
@@ -129,6 +129,8 @@ enum {
 	WIF_ATTR_DVLAN_NAMING,
 	WIF_ATTR_DVLAN_BRIDGE,
 	WIF_ATTR_MIN_HW_MODE,
+	WIF_ATTR_11R_R0KH,
+	WIF_ATTR_11R_R1KH,
 	__WIF_ATTR_MAX,
 };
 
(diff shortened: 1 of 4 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -361,13 +365,27 @@ static int vif_config_security_set(struct blob_buf *b,
...
+
+			/*
+			 * If Radius is configured and Roaming is enabled,
+			 * - disable ft_psk_generate_local. This is required for hostapd to populate R0/R1.
+			 * - populate r0KH and r1KH with broadcast addressing so that the corresponding
+			 * Key Holders are auto-generated
+			 */
+			if (vconf->ft_mobility_domain) {
+				blobmsg_add_bool(b, "ft_psk_generate_local", 0);
+				blobmsg_add_string(b, "r0kh", "ff:ff:ff:ff:ff:ff,*,8a7fcc966ed0691ff2809e1f38c16996");
```
RickSommerville, line 377: What is this magic number  8a7fcc....  ?
What makes use of the content of ft_mobility_domain content?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -1338,7 +1372,6 @@ static int ap_vif_config_set(const struct schema_Wifi_Radio_Config *rconf,
 		if (vconf->ft_psk && vconf->ft_mobility_domain) {
```
RickSommerville, line 1372: You have no change here BUT, for the EAP case, shouldn't we expect vconf->ft_psk =0 and vconf->ft_mobility_domain is a value.  If this were so, then this config will not get applied.

Also this code sets ft_psk_generate_local which then gets rewritten later in vif_config_security_set().  This currently happens in the correct order but seems to be asking for confusion later.

Finally, this answers my earlier question; ft_mobility_domain is consumed her.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
RickSommerville: Removed from code.

RickSommerville (approved):
Looks good.  Removes/resolves previous comments.

RickSommerville:
merged fa5c8b06, thanks

PR #232, opened 2021-04-14
openwrt: Disble wan dhcp renew for topology change

chaitanyagodavarthi:
check for no-renew flag in dhcp.sh before
renewing dhcp lease. Renewing dhcp lease
every time a new interface is added (topology change)
to the wan is causing trouble with some dhcp servers
as they consider receiving renew many times in
a quick succession as a DOS attack. So for now
we disable dhcp lease renew when a new interface
is added.


chaitanyagodavarthi:
Already merged in pending

PR #234, opened 2021-04-19
netifd: Dont renew dhcp on bridge topo change

chaitanyagodavarthi:
Do not renew the dhcp on bridge topology change
as this causes some of the dhcp servers to fail.
This occurs when many interfaces are added on to
a bridge and for every interface there is a dhcp
renew. The server considers receiving renew many
times in a quick succession as a DOS attack. So for
now we disable dhcp lease renew when a new interface
is added.


RickSommerville (approved):

RickSommerville:
Merged b717b136, thank you.

PR #235, opened 2021-04-19
Wifi-2028. WF194C Serial number changes with every upgrade

netexp-ravi:
This commit will fix changing serial number on every sys-upgrade.


RickSommerville (approved):

RickSommerville:
Merged 5d8ce349, thanks

PR #240, opened 2021-04-22
WiFi-1921. Fix for hostapd error logs when configuring 16 SSID

netexp-ravi:
When configuring 16 VLAN, we see hostapd complaining
about no buffer space available while handling receive
events. Hence, increasing the socket buffer memory size
to 512KB.


RickSommerville (approved):

RickSommerville:
Merged b4e20e62, thanks

PR #245, opened 2021-04-26
WiFi-2071. Read model info from manufacturing block

netexp-ravi:
Reading model info from flash. If no info present in flash,
then read it from /tmp/sysinfo/boardname.


RickSommerville (approved):

RickSommerville:
Merged 655d284a, thanks

PR #247, opened 2021-04-26
wifi-2070: Increase CSA count

yashvardhan-netex:
- Channel switch in case of high interference was failing
   when multiple ssids are configured. Increasing CSA count helps
   all the vifs to reserve context and get ready before CSA can be
   finalized.
 - Log when we detect high interference


RickSommerville:
Merged 6021a144, thanks

PR #258, opened 2021-05-07
ipq807x: Enable bridge-mgr to control port mac address learning (roaming fix option B)

ecsv:
The switch in the IPQ807x/IPQ60xx devices will automatically learn the mac addresses behind a port. But it will not unlearn this entry when some mac switches from the ethernet port to the CPU port. This will for example happens when a device roams from on AP to another AP. At least when both are APs are bridging the wifi traffic directly or indirectly (mesh) to the same ethernet broadcast domain.

As result, the roaming device can no longer receive any ethernet packets which the AP is expected to receive on the ethernet port. This state will be kept for a couple of minutes until the entry in the FDB is dropped automatically. But it is still possible for the wifi device to send data via the ethernet during this whole time.

One solution is to just disable learning on all ports. The other option would be to enable the qca bridge-mgr which takes care of gathering the events from the bridge and forwards it to the qca-ssdk (to manipulate the state of the switch). The latter option was chosen to follow the approach which QCA is also using in their QSDK.

This is option B. Option A can be found in #259

### DIFF
--- a/feeds/ipq807x/qca-nss-clients/Makefile
+++ b/feeds/ipq807x/qca-nss-clients/Makefile
@@ -1,0 +1,626 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=qca-nss-clients
+PKG_SOURCE_PROTO:=git
+PKG_BRANCH:=master
+PKG_RELEASE:=2
+PKG_SOURCE_URL:=https://source.codeaurora.org/quic/qsdk/oss/lklm/nss-clients/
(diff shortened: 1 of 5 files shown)


blogic:
@ecsv thanks for looking into this. I guess B is probably the more complex, yet cleaner solution ?!

ecsv:
> @ecsv thanks for looking into this. I guess B is probably the more complex, yet cleaner solution ?!

Correct - despite my overall opinion about QCA developed solutions.

blogic:
> > @ecsv thanks for looking into this. I guess B is probably the more complex, yet cleaner solution ?!
> 
> Correct - despite my overall opinion about QCA developed solutions.

:-)

blogic:
merged, Thanks !

PR #265, opened 2021-05-14
APC: Fix BDR takes place of DR

chaitanyagodavarthi:
Raise the priority of BDR in case
when DR is inactive, so that in the
election process BDR is elected as
DR. 
Also add a new apc state WT, 
this is when its waiting for election.


ArifAlam (approved):

RickSommerville:
Merged acc7a244, thanks

PR #266, opened 2021-05-17
APC: relinquish DR/BDR mode on Cloud conn loss

chaitanyagodavarthi:
If there is a cloud connectivity loss, we do not
want the AP to be elected as a DR/BDR since that would
mean the APs configuration could mismatch with all the
other APs linked to it.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -35,6 +35,7 @@ ovsdb_table_t table_Hotspot20_Icon_Config;
 ovsdb_table_t table_APC_Config;
 ovsdb_table_t table_APC_State;
 unsigned int radproxy_apc = 0;
+extern json_t* ovsdb_table_where(ovsdb_table_t *table, void *record);
 
 static struct uci_package *wireless;
 struct uci_context *uci;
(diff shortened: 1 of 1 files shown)


RickSommerville (changes requested):
Github says this code will not merge cleanly,  please rebase.

Why are you polling the Manager table instead of listening for events?
There is an example in command/src/webserver.c that does a listen on the Manager table.
You can improve on that by using a filter so only get notified of "status" changes; see cm2_ovsdb.c/cm2_ovsdb_init() around line 2087 (this is another example of listen on Manager table).

chaitanyagodavarthi:
> Github says this code will not merge cleanly, please rebase.
> 
> Why are you polling the Manager table instead of listening for events?
> There is an example in command/src/webserver.c that does a listen on the Manager table.
> You can improve on that by using a filter so only get notified of "status" changes; see cm2_ovsdb.c/cm2_ovsdb_init() around line 2087 (this is another example of listen on Manager table).

I have rebased the patch. I did initially use the method you mention above, however it would not work in case where the AP always remains in the is_connected=false state from the beginning and doesn't change state at all.  in which case we would not get any events.

RickSommerville:
Could you handle by assuming "not connected" until notified otherwise?

chaitanyagodavarthi:
> Could you handle by assuming "not connected" until notified otherwise?

Ok, In this case we need a one time timer to poll that "not connected" variable.

RickSommerville (approved):

RickSommerville:
Merged d4oe1f9d, thanks.  WIFI-2343

PR #267, opened 2021-05-17
WIFI-2381: Query DigiCert's API for Redirector's address

brennerm:
Signed-off-by: Max Brenner <xamrennerb@gmail.com>

### DIFF
--- a/feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
+++ b/feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
@@ -1,11 +1,30 @@
 #!/bin/sh
 
-if [ $# -ne 1 ] ; then
-	echo "Usage: $0 <redirector address>" >&2
-	exit 1
+AP_PRIVATE_KEY_FILE="/usr/opensync/certs/client_dec.key"
+AP_CERTIFICATE_FILE="/usr/opensync/certs/client.pem"
+AP_DEVICE_ID_FILE="/usr/opensync/certs/client_deviceid.txt"
+DIGICERT_API_URI="clientauth.one.digicert.com"
+
+if [ "$1" = "-h" ]; then
+  echo "Usage: $0 [redirector address]" >&2
+  exit 1
 fi
 
-redirector_addr=$1
+# Query DigiCert's API if redirector wasn't specified
+if [ -z "$1" ]; then
+  # TODO: this command should be retried if it fails
+  digicert_device_id=`cat ${AP_DEVICE_ID_FILE}`
+  device_data=`curl -s \
+    --key "${AP_PRIVATE_KEY_FILE}" \
+    --cert "${AP_CERTIFICATE_FILE}" \
+    "https://${DIGICERT_API_URI}/iot/api/v2/device/${digicert_device_id}"`
+
+  controller_ip=`echo ${device_data} | jq '.fields | .[] | select ( .name == "Redirector" ) | .value' | tr -d '"'`
+  # TODO: we should get the port with the redirector record and only default to 6643 if no port was specified
+  redirector_addr="ssl:${controller_ip}:6643"
+else
+  redirector_addr=$1
+fi
 
 uci set system.tip.redirector="${redirector_addr}"
 uci commit system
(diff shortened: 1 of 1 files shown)


leonidlm (approved):

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
```diff
@@ -1,11 +1,30 @@
 #!/bin/sh
 
-if [ $# -ne 1 ] ; then
-	echo "Usage: $0 <redirector address>" >&2
-	exit 1
+AP_PRIVATE_KEY_FILE="/usr/opensync/certs/client_dec.key"
+AP_CERTIFICATE_FILE="/usr/opensync/certs/client.pem"
+AP_DEVICE_ID_FILE="/usr/opensync/certs/client_deviceid.txt"
```
RickSommerville, line 5: The DigiCert option depends on the existence of  AP_DEVICE_ID_FILE="/usr/opensync/certs/client_deviceid.txt".
How does this get created?  It does not exist on my CIG WF194c.
Should the code check for the existence of the file and error out instead of doing a call to DigiCert that will fail?

brennerm: This file is part of the AP package that is being created by the PKI scripts as you can see [here](https://github.com/Telecominfraproject/wlan-pki-cert-scripts/blob/digicert/digicert-generate-ap-certs.sh#L72).

Will add a file existence check. :+1:

RickSommerville: Thanks for the link.  Should this other branch also have a PR or be part of this PR?  Once you add the file check, do you want this PR merged alone?

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
```diff
@@ -1,11 +1,35 @@
...
+  fi
+
+  # TODO: this command should be retried if it fails
+  digicert_device_id=`cat ${AP_DEVICE_ID_FILE}`
+  device_data=`curl -s \
+    --key "${AP_PRIVATE_KEY_FILE}" \
+    --cert "${AP_CERTIFICATE_FILE}" \
+    "https://${DIGICERT_API_URI}/iot/api/v2/device/${digicert_device_id}"`
+
+  controller_ip=`echo ${device_data} | jq '.fields | .[] | select ( .name == "Redirector" ) | .value' | tr -d '"'`
```
brennerm, line 27: @RickSommerville Do you know if `jq` is available on a stock firmware image?

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
brennerm: "generate data to be loaded to the AP" <- this

We need to wait for the merge in the PKI scripts repo before we can merge this PR.

feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
RickSommerville: I don't think so.  I tried to just execute 'jq' from SSH to AP and it reports "-ash: jq: not found".
"which jq" similarly reports nothing.

brennerm: Mmh do you think it's feasible to include it in the image? Otherwise we will have trouble to reliably extract the redirector URL from the Digicert response.

brennerm: Or maybe you know of any other tool that is already installed and capable of handling JSON.

RickSommerville: Look at wan-ap-config/etc/uci-defaults/90-tip-hostname.  It uses "jsonfilter".  Relevant code is:

MAC=$(cat /etc/board.json | jsonfilter -e '@["network"]["wan"]["macaddr"]' | tr -d :)

brennerm: Awesome, thanks! Replaced the query using jsonfilter.

brennerm:
@RickSommerville Will this script get executed automatically on boot?

RickSommerville:
Merged to pending a1ff0f90

PR #268, opened 2021-05-17
Wifi-2153. Fix for STA Assoc rejection with reason code 17

netexp-ravi:
If a previously associated-and-authorized STA comes for a new
connection (Assoc Request), AP rejects the Assoc with code 17.
Solution is to delete the STA entry at AP before re-adding it. 
This clears out the state-machine at the AP and
accepts the new connection from the STA.


RickSommerville:
Merged d3695266, thanks

PR #269, opened 2021-05-18
WiFi-1827. Updates for RRM Channel Hop based on Noise Floor threshold

RickSommerville:
Average Noise Floor is calcluated for the last few samples based on the
configured Noise Floor time. This calculated avg Noise Floor is compared
against the configured Noise Floor threshold to decide if a Channel hop
is required. While calculating the average, each noise sample is first
converted from dBm to milliWatts, an avaerage of the all the noise
samples in milliWatt is calculated, then the average is converted back
to dBm.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
@@ -9,8 +9,6 @@
 #include "uci.h"
 #include "utils.h"
 
-#define RRM_CHANNEL_INTERVAL     15.0
-
 struct blob_buf b = { };
 struct blob_buf del = { };
 struct uci_context *uci;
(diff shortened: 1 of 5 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
```diff
@@ -103,6 +101,56 @@ void get_channel_bandwidth(const char* htmode, int *channel_bandwidth)
...
+	int ii;
+	double avg_mW = 0;
+	int num_samples = config_time/sample_time;
+
+	if (num_samples >= RRM_MAX_NF_SAMPLES)
+		num_samples = RRM_MAX_NF_SAMPLES;
+
+	/* Replace the oldest element in the list with the new element */
+	rrm_data->rrm_chan_nf_samples[rrm_data->rrm_chan_nf_next_el] = nf;
+
```
RickSommerville, line 126: I propose converting to mW before storing in the array.  Then the calculation is easier and we just convert on the return value.

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
```diff
@@ -103,6 +101,56 @@ void get_channel_bandwidth(const char* htmode, int *channel_bandwidth)
...
+		return 0;
+	}
+
+	/*
+	 * Convert dBm to milliWatts, calculate average,
+	 * convert the averaged milliWats back to dBm
+	 */
+	for (ii = 0; ii < num_samples; ii++)
+	{
+		avg_mW += dBm_to_mwatts(rrm_data->rrm_chan_nf_samples[ii]);
```
RickSommerville, line 142: Then this would just be addition.

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
```diff
@@ -103,6 +101,56 @@ void get_channel_bandwidth(const char* htmode, int *channel_bandwidth)
...
+	}
+	avg_mW = avg_mW/num_samples;
+
+	return ((int)(mWatts_to_dBm(avg_mW)));
```
RickSommerville, line 146: And this would be unchanged.

Or we could convert the threshold 1 time and just make the decision in mW instead of dBm

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
```diff
@@ -122,58 +169,52 @@ void rrm_nf_timer_handler(struct ev_loop *loop, ev_timer *timer, int revents)
...
 		rrm_config = rrm_get_rrm_config(radio->config.type);
 
 		if (rrm_config == NULL)
 			continue;
 
-		if (nf < rrm_config->rrm_data.noise_lwm )
-		{
-			rrm_config->rrm_data.noise_lwm = nf;
-			LOGD("[%s] noise_lwm set to %d", radio->config.if_name, nf);
+		if (rrm_config->rrm_data.backup_channel == 0)
```
RickSommerville, line 186: Do we also need a test to make sure we are not already on the backup channel?

I see at line 228 that we reset the data after a channel switch, but we could still accumulate new data and try to switch again.

RickSommerville (commented):
PR #254 was against 'trunk', I created this one against pending.

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
netexp-ravi: Ok. I will make the change.

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
netexp-ravi: OK.

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
netexp-ravi: I would vote for leaving this as it is. This way the input and output of this function stay consistent (dBm).

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
netexp-ravi: Good point. Thanks for bringing this up.  This check might be required.
I was thinking to not even collect the nf samples if the configured channel is same as backup channel.
So, we can add a check at the top of this function like below,

		if ((rrm_config->rrm_data.backup_channel == 0) || 
                     (rrm_config->rrm_data.backup_channel == ConfiguredChannel))
			continue;

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
RickSommerville: Makes sense.

feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
netexp-ravi: done

RickSommerville:
Merged 7569b618, thanks

PR #271, opened 2021-05-18
ucentral: Don't use batmand

ecsv:
batman-adv is a layer 2 implementation of the B.A.T.M.A.N. (IV/V) algorithm. The batmand is an old layer 3 implementation of the B.A.T.M.A.N. III algorithm. The latter is not really useful when already using batman-adv on top of the IBSS/meshpoint interfaces. And running layer 3 B.A.T.M.A.N. III on top of layer 2 B.A.T.M.A.N. IV might lead to all kind of unexpected behavior.

----

Maybe it is actually useful in your configuration. But this kind of installation often happens when people didn't notice that batmand is not neeeded/wanted when using batman-adv. So I just wanted to point this out.


blogic:
@ecsv thanks !

ecsv:
Seems like this was applied in 6621537977b4c93d8e117aeaccd0cc46d5af1309

PR #275, opened 2021-05-19
wifi-ax: Update batman-adv to openwrt-routing 21.02

ecsv:
The batman-adv for non-802.11ax devices was a completely different one than for 802.11ax devices. The latter is using a private copy because the QCA kernel is extremely old and doesn't match the kernel from OpenWrt 21.02. And the mac80211 version (and its backports/compat code) is not fully compatible with the official one.

At least import the newest version from openwrt-routing openwrt-21.02 and change to build against the special kernel and mac80211 version.

The used commit from openwrt-routing is 5d09821ed7c4 ("batman-adv: Always send iface index+name in genlmsg")


blogic:
merged, Thanks !

PR #276, opened 2021-05-19
WIFI-2379: fix radius proxy config

ArifAlam:
Remove stale radius proxy config on ovsdb delete.

WIFI-2377: check radius proxy service on APC state change

Start or stop radius proxy service as needed on APC state change.

Radius proxy service is only started if:
- AP is elected as a DR (Designated Router) via APC
- radproxy is set to true in vif config custom options for any
  ssid


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -742,6 +742,10 @@ static void callback_APC_State(ovsdb_update_monitor_t *mon,
 		radproxy_apc = 0;
 		system("ubus call service event '{\"type\": \"config.change\", \"data\": { \"package\": \"wireless\" }}'");
 	}
+
+	/* APC changed: start / stop radius proxy service if needed */
+	vif_check_radius_proxy();
```
RickSommerville, line 747: Does the order of "starting the raised proxy if necessary" and the bus call at line 743 matter?

RickSommerville (commented):
Looks good.  One question.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
ArifAlam: I don't think it matters but I prefer to do the check at the very end.

RickSommerville:
Merged 915041e4, thanks

PR #280, opened 2021-05-20
WIFI-721: Added channel_max_power map to radio state table

shimmy568:
For this PR I added a new field to the Wifi_Radio_State table and added the code to populate the map.

The key for the map represents the number for the channel and the value is the max value that the power can be set to.

This is what the Wifi_Radio_State table looks like with the change in place (EA8300):
image

And this is the same as above but just for an EX227:
image

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -181,6 +181,36 @@ static void radio_state_custom_options_get(struct schema_Wifi_Radio_State *rstat
 	}
 }
 
+static void set_channel_max_power(struct schema_Wifi_Radio_State *rstate,
+                                    int *index, int channel,
+                                    int max_power)
+{
+	rstate->channel_max_power_keys[*index] = channel;
+	rstate->channel_max_power[*index] = max_power;
+	*index += 1;
+	rstate->channel_max_power_len = *index;
+}
+
+// Update the entire channel_max_power map in the radio state
+static void update_channel_max_power(char* phy, struct schema_Wifi_Radio_State *rstate) {
+	int channels[64];
+	int channel_count = phy_get_channels(phy, channels);
+	channel_count += phy_get_dfs_channels(phy, channels + channel_count);
+
+	// Clear the data previously stored in channel_max_power
+	memset(rstate->channel_max_power_keys, 0, sizeof(rstate->channel_max_power_keys));
+	memset(rstate->channel_max_power, 0, sizeof(rstate->channel_max_power));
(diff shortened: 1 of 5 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -181,6 +181,35 @@ static void radio_state_custom_options_get(struct schema_Wifi_Radio_State *rstat
...
+	rstate->channel_max_power_len = *index;
+}
+
+// Update the entire channel_max_power map in the radio state
+static void update_channel_max_power(char* phy, struct schema_Wifi_Radio_State *rstate) {
+	int channels[64];
+	int channel_count = phy_get_channels(phy, channels);
+	
+	// Clear the data previously stored in channel_max_power
+	memset(rstate->channel_max_power_keys, 0, sizeof(rstate->channel_max_power_keys));
```
RickSommerville, line 200: How do we know how many channels this structure can hold?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -181,6 +181,35 @@ static void radio_state_custom_options_get(struct schema_Wifi_Radio_State *rstat
...
+	memset(rstate->channel_max_power, 0, sizeof(rstate->channel_max_power));
+	rstate->channel_max_power_len = 0;
+
+	// Set the new values for channel_max_power
+	int i, index = 0;
+	for (i = 0; i < channel_count; i++) {
```
RickSommerville, line 206: This loop runs over the number of channels we have without any test of the size of the destination buffer.
It may have been fine in your testing and break in some condition where lots of channels are available.

RickSommerville (commented):
In the example output I don't see any DFS channels listed.
They are not technically 'available' because a DFS "channel availability check" has not been done (or not recently enough).  However, we should still report the max_power.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
shimmy568: After looking into the code for the schema generation it should be max 64 items

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
shimmy568: Added check to `i` so it would max out at 64

RickSommerville (approved):
Thanks for adding the DFS channels.

RickSommerville:
Merged e6bb6f7d, thanks.

PR #281, opened 2021-05-20
wifi-2069 Fix for vifC not getting applied to vifS

yashvardhan-netex:
- WM was getting stuck on recvmsg() while receiving netlink event notifications
   Since we are already using event loop to read data as and when it arrives, there was
   no need to keep nl_recvmsgs() in blocking mode for notifications.
   Setting socket to non-blocking mode resolves the issue.
 - Also removed a redundant reload_config call in captive.c.


RickSommerville (approved):

RickSommerville:
Merged a71436d2, thanks.

PR #283, opened 2021-05-22
WIFI-2416  Automatically DigiCert if no redirector setup

RickSommerville:

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/rootfs/common/etc/init.d/opensync
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/rootfs/common/etc/init.d/opensync
@@ -37,6 +37,14 @@ start_service() {
     echo "Setting certificates"
     mkdir -p ${CERTS_DEST_PATH}
     cp ${CERTS_SRC_PATH}/* ${CERTS_DEST_PATH}/
+    echo "Checking Redirector"
+    redirector=$(uci get system.tip.redirector)
+    if [ -z "$redirector" ]; then
+        logger -t opensync "Contacting DigiCert for redirector address"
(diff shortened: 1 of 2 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/rootfs/common/etc/init.d/opensync
```diff
@@ -37,6 +37,14 @@ start_service() {
     echo "Setting certificates"
     mkdir -p ${CERTS_DEST_PATH}
     cp ${CERTS_SRC_PATH}/* ${CERTS_DEST_PATH}/
+    echo "Checking Redirector"
+    redirector=$(uci get system.tip.redirector)
+    if [ -z "$redirector" ]; then
+        logger -t opensync "Contacting DigiCert for redirector address"
+        wlan_ap_redirector.sh
```
chaitanyagodavarthi, line 44: Should we put a full path here?

chaitanyagodavarthi (commented):
Should we put a full path here? and check the exit condition if it failed, then put some default redirector address.

feeds/wlan-ap/opensync/src/platform/openwrt/rootfs/common/etc/init.d/opensync
RickSommerville: There is a default redirector (to an old test server) already in the OpenSync.
The decision was taken earlier not to put a default to a real server because a) the certificates wouldn't match anyway (unless a non-unique certificate was also set as a default) and b) there would be an cost if many AP's connected to a server by default.

RickSommerville:
Merged d7b413bd.

PR #292, opened 2021-05-31
WIFI-1845: Updated rotation configuration file with new date format

shimmy568:
We were experiencing an issue where the date format for the logs would collide with each other if we executed the rotation more than once per day. I've updated the date format to be unix epoch instead removing this issue. I also updated the wildcard used so it wouldn't try and rotate the rotated logs.


RickSommerville (approved):

RickSommerville:
Merged 60bff767, thanks

PR #295, opened 2021-06-01
Wifi-2384. Fix for StatsManager crash

netexp-ravi:
Could not root-cause the issue for SM crash, but there were enough
indications of some kind of memory corruption. Different core captures
show the code crashing at different places. An experiment to use 
unblocking SM socket on the receive end helped avoid the corruption and
subsequently the crashes.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
@@ -37,8 +37,8 @@
 #include "radio.h"
 
 extern struct ev_loop *wifihal_evloop;
-static int nl80211_scan_started;
-static struct unl unl;
+static struct unl unl_req;
+static struct unl unl_notify;
 static ev_io unl_io;
 
 struct nl80211_scan {
@@ -345,7 +345,7 @@ static struct nl_msg *nl80211_call_phy(char *name, int cmd, bool dump)
 	if (idx < 0)
 		return NULL;
 
-	msg = unl_genl_msg(&unl, cmd, dump);
+	msg = unl_genl_msg(&unl_req, cmd, dump);
 	nla_put_u32(msg, NL80211_ATTR_WIPHY, idx);
 
 	return msg;
@@ -461,7 +461,7 @@ static void nl80211_ev(struct ev_loop *ev, struct ev_io *io, int event)
 	nl_cb_set(cb, NL_CB_FINISH, NL_CB_CUSTOM, finish_handler, NULL);
 	nl_cb_set(cb, NL_CB_SEQ_CHECK, NL_CB_CUSTOM, no_seq_check, NULL);
 	nl_cb_set(cb, NL_CB_VALID, NL_CB_CUSTOM, nl80211_recv, NULL);
-	nl_recvmsgs(unl.sock, cb);
+	nl_recvmsgs(unl_notify.sock, cb);
 	nl_cb_put(cb);
 }
 
@@ -470,18 +470,10 @@ static struct nl_msg *nl80211_call_vif(struct nl_call_param *nl_call_param, int 
 	int idx = if_nametoindex(nl_call_param->ifname);
 	struct nl_msg *msg;
 
-	if (!nl80211_scan_started) {
-		unl_genl_subscribe(&unl, "scan");
-
-		ev_io_init(&unl_io, nl80211_ev, unl.sock->s_fd, EV_READ);
-		ev_io_start(wifihal_evloop, &unl_io);
-		nl80211_scan_started = 1;
-	}
-
 	if (!idx)
 		return NULL;
 
-	msg = unl_genl_msg(&unl, cmd, dump);
+	msg = unl_genl_msg(&unl_req, cmd, dump);
 	nla_put_u32(msg, NL80211_ATTR_IFINDEX, idx);
 
 	return msg;
@@ -494,7 +486,7 @@ int nl80211_get_tx_chainmask(char *name, unsigned int *mask)
 	if (!msg)
 		return -1;
 
-	return unl_genl_request(&unl, msg, nl80211_chainmask_recv, mask);
+	return unl_genl_request(&unl_req, msg, nl80211_chainmask_recv, mask);
 }
 
 int nl80211_get_oper_channel(char *name, unsigned int *chan)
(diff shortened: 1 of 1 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
```diff
@@ -597,18 +598,26 @@ int nl80211_scan_dump(struct nl_call_param *nl_call_param)
...
+	if (unl_genl_init(&unl_notify, "nl80211") < 0) {
+		LOGE("failed to spawn nl80211");
+		return -1;
+	}
+
+	if (nl_socket_set_buffer_size(unl_notify.sock, 262144, 0) < 0)
 		LOGE("stats_nl80211: Failed to set nl socket buffer size");
 
+	if (nl_socket_set_nonblocking(unl_notify.sock))
+		LOGE("radio_nl80211: Failed to set stats nl socket in the non blocking mode");
```
yashvardhan-netex, line 620: nitpick, File name is stats_nl80211 .

yashvardhan-netex (commented):
What all tests were done? Did we check if all the reports are sent up to the cloud as expected?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
netexp-ravi: Thanks. I will fix this.

netexp-ravi:
Did not perform any specific tests. 
We used to see SM crashes as the AP is left up for long hours. With the changes, we saw no SM crash while the wifi6 AP stayed up for more than 4 days. And no SM crash with a wifi5 AP up for about 24 hours.
The Cloud UI does show up all the graphs populated, so I assume the AP sends all the reports that it is supposed to.
We still see choppy graphs though, and that is analyzed to be a different issue between the Cloud and the Gateway.

yashvardhan-netex:
> Did not perform any specific tests.
> We used to see SM crashes as the AP is left up for long hours. With the changes, we saw no SM crash while the wifi6 AP stayed up for more than 4 days. And no SM crash with a wifi5 AP up for about 24 hours.
> The Cloud UI does show up all the graphs populated, so I assume the AP sends all the reports that it is supposed to.
> We still see choppy graphs though, and that is analyzed to be a different issue between the Cloud and the Gateway.

You may want to have a quick look at the metrics and their contents on UI to see if it is what is sent by AP and is what is expected.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
```diff
@@ -471,17 +472,17 @@ static struct nl_msg *nl80211_call_vif(struct nl_call_param *nl_call_param, int
 	struct nl_msg *msg;
 
 	if (!nl80211_scan_started) {
```
RickSommerville, line 474: I've always wondered about this block of code.  It looks to me like it could be done at start-up.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
```diff
@@ -597,18 +598,26 @@ int nl80211_scan_dump(struct nl_call_param *nl_call_param)
...
 		return -1;
 
-	return unl_genl_request(&unl, msg, nl80211_scan_dump_recv, nl_call_param);
+	return unl_genl_request(&unl_req, msg, nl80211_scan_dump_recv, nl_call_param);
 }
 
 int stats_nl80211_init(void)
 {
-	if (unl_genl_init(&unl, "nl80211") < 0) {
+	if (unl_genl_init(&unl_req, "nl80211") < 0) {
```
RickSommerville, line 606: Should unl_req also have the increased buffer size?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
netexp-ravi: I agree. 
Logically, it does not make any difference as we are gating the initialization with "nl80211_scan_started". But yes, we can move this block to init and get rid of "nl80211_scan_started".
I will make this change unless anyone objects to it.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
yashvardhan-netex: I believe it is not needed. unl_req is used for netlink request and receive messages as opposed to unl_notify which is used to receive notifications. It is the notifications buffer that could grow if there is a delay in reading or if we have a burst of events as seen with connecting multiple clients at one go.

netexp-ravi:
> > Did not perform any specific tests.
> > We used to see SM crashes as the AP is left up for long hours. With the changes, we saw no SM crash while the wifi6 AP stayed up for more than 4 days. And no SM crash with a wifi5 AP up for about 24 hours.
> > The Cloud UI does show up all the graphs populated, so I assume the AP sends all the reports that it is supposed to.
> > We still see choppy graphs though, and that is analyzed to be a different issue between the Cloud and the Gateway.
> 
> You may want to have a quick look at the metrics and their contents on UI to see if it is what is sent by AP and is what is expected.

I checked the OpenSync Gateway logs, and they do contain the Client report and ON-Chan report sent from the AP. Also verified for the content in the ON-Chan report, they match between the AP and the Gateway.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
netexp-ravi: Done

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats_nl80211.c
netexp-ravi: Done

yashvardhan-netex:
Looks good.

RickSommerville (approved):

RickSommerville:
Merged 89640188, thanks.

PR #296, opened 2021-06-02
WIFI-2517: auto config radius params when radius proxy is enabled

ArifAlam:
Auto configure per-ssid radius params to talk to the radius proxy service.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
@@ -388,7 +388,8 @@ static int vif_config_security_set(struct blob_buf *b,
 	const char *mode = SCHEMA_KEY_VAL(vconf->security, SCHEMA_CONSTS_SECURITY_MODE);
 	unsigned int i;
 	unsigned int acct_interval;
-	const char *auth_server, *auth_port, *auth_secret, *security_key, *acct_server;
+	const char *auth_server, *auth_port, *auth_secret, *security_key;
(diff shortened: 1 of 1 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -430,16 +431,22 @@ static int vif_config_security_set(struct blob_buf *b,
 					auth_server = local_server;
 					acct_server = local_server;
 				}
+				auth_port = "1812";
+				auth_secret = "secret";
+				acct_port = "1813";
+				acct_secret = "secret";
```
RickSommerville, line 437: Code seems to make sense but is the secret really "secret"?
Doesn't this imply there is a RADIUS server somewhere with secret == "secret"?
Seems like a bad idea.

ArifAlam: This is the communication between hostapd which is running inside DR/BDR/OR and the proxy service running inside the DR.  The radius endpoint here is the proxy service.

ArifAlam: In other words, the server ip will always point to the entity running the radius proxy service. In case of DR it'll be 127.0.0.1.  In case of OR/BDR it will be the IP address of the DR.

RickSommerville (approved):

RickSommerville:
Merged f68851cc, thanks.

PR #297, opened 2021-06-02
wifi-2512 APC state gets stuck in WT mode on bootup

chaitanyagodavarthi:
Sometimes the APC_State mode gets stuck in WT mode
on bootup. Sometimes the APC State gets updated at
the same time as disabling the APC (SIGTERM puts it in NC mode)
So it may get updated to WT just after SIGTERM handler
puts it in NC mode.
Fix this by setting APC_State in NC mode again after disabling
the APC from opensync (wm).

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -814,13 +814,38 @@ static int conn_since = 0;
 
 static void apc_enable(bool flag) {
 
-	SCHEMA_SET_INT(apc_conf.enabled, flag);
-	if (!ovsdb_table_update(&table_APC_Config, &apc_conf)) {
-		LOG(ERR, "%s:APC_Config: failed to update", __func__);
-		return;
-	}
+	struct schema_APC_State apc_state;
+	json_t *where;
+	
 	LOGI("APC %s: %s APC", __func__, flag?"enable":"disable");
+	if (flag == false) {
+		where = ovsdb_table_where(&table_APC_State, &apc_state);
+		if (false == ovsdb_table_select_one_where(&table_APC_State,
+						  where, &apc_state)) {
+			LOG(ERR, "%s: APC_State read failed", __func__);
+			apc_state.enabled = true;
+		}
 
+		if (apc_state.enabled == true) {
+			SCHEMA_SET_INT(apc_conf.enabled, flag);
+			if (!ovsdb_table_update(&table_APC_Config, &apc_conf)) {
+				LOG(ERR, "%s:APC_Config: failed to update", __func__);
+				return;
+			}
+			SCHEMA_SET_STR(apc_state.mode, "NC");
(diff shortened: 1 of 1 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -847,6 +872,7 @@ apc_cld_mon_cb(struct schema_Manager *mgr)
 		LOGI("The return value: %d\n", WEXITSTATUS(ret));
 		link = WEXITSTATUS(ret);
 		if (link == 0) {
+			LOGI("APC link down");
```
RickSommerville, line 875: More info, lower priority.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -847,6 +872,7 @@ apc_cld_mon_cb(struct schema_Manager *mgr)
 		LOGI("The return value: %d\n", WEXITSTATUS(ret));
```
RickSommerville, line 872: I see this log a lot on my AP.  It needs to be more informative and lower priority (like DEBUG or TRACE).

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
chaitanyagodavarthi: done

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
chaitanyagodavarthi: done

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -814,13 +814,38 @@ static int conn_since = 0;
...
-		LOG(ERR, "%s:APC_Config: failed to update", __func__);
-		return;
-	}
+	struct schema_APC_State apc_state;
+	json_t *where;
+	
 	LOGI("APC %s: %s APC", __func__, flag?"enable":"disable");
+	if (flag == false) {
+		where = ovsdb_table_where(&table_APC_State, &apc_state);
+		if (false == ovsdb_table_select_one_where(&table_APC_State,
```
NagConnectus, line 823: will it be good to make  **flag** == ovsdb_table_select_one_where(&amp;table_APC_State, where, &amp;apc_state)), since flag will be always false within this if condition.

chaitanyagodavarthi: That flag represents if the APC is to be enabled/disabled, so it would be confusing if we use it to check return value of that function.

NagConnectus (approved):

RickSommerville (approved):

RickSommerville:
Merged b71059cc, thanks

PR #298, opened 2021-06-02
Wifi-2529. SM crash with signal 11 when DM tries to re-spawn the SM p…

netexp-ravi:
Wifi-2529. SM crash with signal 11 when DM tries to re-spawn the SM process.

The earlier changes for detaching the nl sockets for SM activity
also included a small change to relocate the event subscription.
This relocation of code is causing the crash due to possible
NULL pointer dereference.
Reverting the event subscription relocation code to avoid the
continuous SM crash.


yashvardhan-netex (approved):

NagConnectus (approved):

blogic:
merged, Thanks !

PR #299, opened 2021-06-03
Wifi-2529. SM crash when DM tries to re-spawn the SM process.

netexp-ravi:
The earlier changes for detaching the nl sockets for SM activity
also included a small change to relocate the event subscription.
This relocation of code is causing the crash due to possible
NULL pointer dereference.
Reverting the event subscription relocation code to avoid the
continuous SM crash.


RickSommerville (approved):
Looks good

RickSommerville:
Merged, e528b832, thanks

PR #301, opened 2021-06-04
WIFI-2410: Added patch to host-apd to fix ubus assoc validation

shimmy568:

Fairly small change. Just moved the code to send the assoc response to be after the ubus validation code runs instead of before it. I was unable to test this code due to the fact that the auth response also checks rrsi levels as well as the WiFi client I was using which caused it to be impossible to test (at least with my knowledge). 

But it shouldn't break anything, just possibly prevent weird edge cases with roaming

### DIFF
--- a/feeds/wifi-ax/hostapd/files/hostapd.sh
+++ b/feeds/wifi-ax/hostapd/files/hostapd.sh
@@ -214,6 +214,9 @@ hostapd_common_add_bss_config() {
 	config_add_int maxassoc max_inactivity
 	config_add_boolean disassoc_low_ack isolate short_preamble
 
+config_add_int signal_connect signal_stay signal_poll_time \
+		signal_drop_reason signal_strikes
+
 	config_add_int \
 		wep_rekey eap_reauth_period \
 		wpa_group_rekey wpa_pair_rekey wpa_master_rekey
@@ -485,6 +488,7 @@ hostapd_set_bss_options() {
 	local wep_rekey wpa_group_rekey wpa_pair_rekey wpa_master_rekey wpa_key_mgmt
 
 	json_get_vars \
+	signal_connect signal_stay signal_poll_time signal_drop_reason signal_strikes \
 		wep_rekey wpa_group_rekey wpa_pair_rekey wpa_master_rekey wpa_strict_rekey \
 		wpa_disable_eapol_key_retries tdls_prohibit \
 		maxassoc max_inactivity disassoc_low_ack isolate auth_cache \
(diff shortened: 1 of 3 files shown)


RickSommerville (approved):
Looks good.

feeds/wifi-ax/hostapd/patches/601-fix_ubus_assoc.patch
```diff
@@ -0,0 +1,26 @@
...
+-		reply_res = send_assoc_resp(hapd, sta, mgmt->sa, resp, reassoc,
+-					    pos, left, rssi, omit_rsnxe);
+-	os_free(tmp);
+-
+ 	ubus_resp = hostapd_ubus_handle_event(hapd, &req);
+ 	if (ubus_resp) {
+ 		wpa_printf(MSG_DEBUG, "Station " MACSTR " assoc rejected by ubus handler.\n",
+@@ -4576,6 +4571,11 @@ static void handle_assoc(struct hostapd_
+ 		goto fail;
+ 	}
```
yashvardhan-netex, line 17: Can we move this even up? I mean why do the other stuff if we are not going to associate the client in case of low rssi.

feeds/wifi-ax/hostapd/patches/601-fix_ubus_assoc.patch
```diff
@@ -0,0 +1,26 @@
+--- a/src/ap/ieee802_11.c
```
yashvardhan-netex, line 1: I believe we should apply the patch on top of the latest patch. Sometimes the hunk succeeds but it might corrupt the other patches.

feeds/wifi-trunk/hostapd/patches/601-fix_ubus_assoc.patch
```diff
@@ -0,0 +1,26 @@
+--- a/src/ap/ieee802_11.c
```
yashvardhan-netex, line 1: Same here. Please apply on top of the latest patch.

yashvardhan-netex (commented):
I think we should still find a way to test it. Nevertheless, I hope we have tested the pass scenario, I mean testing the association with multiple clients. You may use lanforge for the same.

feeds/wifi-ax/hostapd/files/hostapd.sh
```diff
@@ -214,6 +214,9 @@ hostapd_common_add_bss_config() {
 	config_add_int maxassoc max_inactivity
 	config_add_boolean disassoc_low_ack isolate short_preamble
 
+config_add_int signal_connect signal_stay signal_poll_time \
```
shimmy568, line 217: Ran into an issue on the wifi-6 ap's where we weren't applying the configuration for `signal_stay` and some other related fields to the hostapd configuration file. Making the client_disconnect_threshold field not work for wifi-6 aps.

RickSommerville:
Merged 8a697710, thanks

PR #302, opened 2021-06-07
Wifi-2542. Fix for populating correct ID in AWLAN_Node

netexp-ravi:
This fix applies to EA8300, WF194c and EAP102. For these APs,
the br-wan is mapped to eth1 instead of eth0. Adding fix to
use the mac address from eth1 as ID for these particular APs.


RickSommerville (approved):

RickSommerville:
Merged c0cb6baa, thanks.

PR #305, opened 2021-06-10
add firmware upload to S3

brennerm:


blogic:
merged into uCentral-trunk, Thanks !

PR #307, opened 2021-06-10
WiFi-2594: Block lan ssh when deployed

shimmy568:
To prevent someone sshing into one of our AP's in the field we need to add some kind of firewall rule to block the port. I've made it so that when it's not connected to the cloud the ssh port is open to allow for debugging and first time setup

The script I added may be a little overkill but I needed a way to modify the existing firewall rule without creating a new one and it felt a little awkward to do it in C and then also duplicate the logic in the redirector script.

### DIFF
--- a/feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
+++ b/feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+if [ "$#" -ne 1 ]; then
+    echo "Illegal number of parameters"
+    exit 1
+fi
+
+action=$1
+
+[ "$action" = "enable" ] && [ ! -f /etc/ssh-flag ] && uci set firewall.lan_ssh_rule.target='REJECT'
+[ "$action" = "disable" ] && uci set firewall.lan_ssh_rule.target='ACCEPT'
+uci commit firewall
+/etc/init.d/firewall restart
+
+exit 0
--- a/feeds/wlan-ap/wlan-ap-config/files/etc/uci-defaults/99-tip-network
+++ b/feeds/wlan-ap/wlan-ap-config/files/etc/uci-defaults/99-tip-network
@@ -6,4 +6,15 @@ uci set network.wan.metric=1
 uci set network.lan.metric=10
 uci set network.wan.vlan_filtering=1
 uci set network.lan.vlan_filtering=1
+
+# Add LAN ssh rule
+uci set firewall.lan_ssh_rule=rule
+uci set firewall.lan_ssh_rule.src='lan'
+uci set firewall.lan_ssh_rule.target='ACCEPT'
+uci set firewall.lan_ssh_rule.proto='tcp'
+uci set firewall.lan_ssh_rule.dest_port='22'
+uci set firewall.lan_ssh_rule.name="Lan SSH Rule"
+uci commit firewall
+/etc/init.d/firewall restart
+
 exit 0
--- a/feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
+++ b/feeds/wlan-ap/opensync/files/bin/wlan_ap_redirector.sh
@@ -40,6 +40,9 @@ else
   redirector_addr=$1
 fi
 
+# Enable lan ssh accsess
+lan-ssh-firewall disable
+
 echo "${redirector_addr}" > /usr/opensync/certs/redirector.txt
 /etc/init.d/uhttpd enable
 /etc/init.d/uhttpd start
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/command/src/webserver.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/command/src/webserver.c
@@ -10,6 +10,13 @@
 static ovsdb_table_t table_Manager;
 static int done;
 
+
+static void disable_ssh() 
+{
+	LOGN("Disabling LAN ssh access");
+	system("lan-ssh-firewall enable");
+}
+
 static void stop_http(void *arg)
 {
 	LOGN("Stopping webserver");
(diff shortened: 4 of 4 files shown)


feeds/wlan-ap/wlan-ap-config/files/etc/uci-defaults/99-tip-data
```diff
@@ -170,3 +170,16 @@ uci set system.tip.id="${ID}"
 uci set system.tip.manuf_abbr="${MANUF_ABBR}"
 
 uci commit
+
```
RickSommerville, line 173: This would be better in 99-tip-network instead of 99-tip-data.

RickSommerville (commented):
No test cases in the defect WIFI-2594.  
Specific concern:  can SSH pass through the AP to an  upstream system?  

Looks like we can disable this for test purposes by manually flipping the new setting from target=REJECT to target=ACCEPT.  If this is done it will be stable in this state.  Correct?

shimmy568:
> Looks like we can disable this for test purposes by manually flipping the new setting from target=REJECT to target=ACCEPT. If this is done it will be stable in this state. Correct?

@RickSommerville Yes as long as the redirector script isn't called again which would reset the value back to REJECT.

Opening up the ssh port should just be as simple as running `lan-ssh-firewall disable`

feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+if [ "$#" -ne 1 ]; then
+    echo "Illegal number of parameters"
+    exit 1
+fi
+
+action=$1
+
+index=$(uci show firewall | grep "Lan SSH Rule" | cut -c 16-17)
```
RickSommerville, line 10: What if no rule?  Add the rule in ACCEPT state.

feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
```diff
@@ -0,0 +1,17 @@
...
+
+[ "$action" = "enable" ] && uci set firewall.@rule["$index"].target='REJECT'
```
RickSommerville, line 12: Make REJECT depend on some "developer" flag file in /etc so that we can work easier locally.

shimmy568:
Test cases: https://telecominfraproject.atlassian.net/browse/WIFI-2594?focusedCommentId=14709

I also tested to see if the default firewall (uci default script) rule gets added in after a basic sysupgrade. It **does** appear to get run so we shouldn't have any issues where the firewall rule doesn't exist and we try and modify it.

feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
shimmy568: I tested to see if the default firewall (uci default script) rule gets added in after a basic sysupgrade. It does appear to get run so we shouldn't have any issues where the firewall rule doesn't exist and we try and modify it.

feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
shimmy568: Added a check for a `/etc/ssh-flag` file before we set the rule to `REJECT`

feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
```diff
@@ -0,0 +1,17 @@
...
+    echo "Illegal number of parameters"
+    exit 1
+fi
+
+action=$1
+
+index=$(uci show firewall | grep "Lan SSH Rule" | cut -c 16-17)
+
+[ "$action" = "enable" ] && [ ! -f /etc/ssh-flag ] && uci set firewall.@rule["$index"].target='REJECT'
+[ "$action" = "disable" ] && uci set firewall.@rule["$index"].target='ACCEPT'
```
RickSommerville, line 13: Note A.  These lines become:
[ "$action" = "enable" ] && [ ! -f /etc/ssh-flag ] && uci set firewall.lan_ssh_rule.target='REJECT'
[ "$action" = "disable" ] && uci set firewall.lan_ssh_rule.target='ACCEPT'

feeds/wlan-ap/wlan-ap-config/files/etc/uci-defaults/99-tip-network
```diff
@@ -6,4 +6,18 @@ uci set network.wan.metric=1
 uci set network.lan.metric=10
 uci set network.wan.vlan_filtering=1
 uci set network.lan.vlan_filtering=1
+
+# Only add firewall rule if it doesn't already exist
+if ! uci show firewall | grep -q "Lan SSH Rule"
+then
+	uci add firewall rule
+	uci set firewall.@rule[-1].src='lan'
```
RickSommerville, line 14: Note B:

Line 13 becomes:   uci set firewall.rule='lan_ssh_rule'
Line 14 and on become:uci set firewall.lan_ssh_rule.<option>=<value>

RickSommerville (changes requested):

feeds/wlan-ap/opensync/files/bin/lan-ssh-firewall
RickSommerville: Cutting the 16 and 17th characters from the uci output is risky.  What if it is a 1 digit number?  And I am still worried that if the section is not present the script will fail or will modify some other section.

feeds/wlan-ap/wlan-ap-config/files/etc/uci-defaults/99-tip-network
yashvardhan-netex: Minor change: Line 13 would be uci set firewall.lan_ssh_rule=rule

RickSommerville (approved):
Looks good, thanks.
Please reply with the result of Remote SSH test.

yashvardhan-netex:
> Looks good, thanks.
> Please reply with the result of Remote SSH test.

Tested, Remote ssh works fine after lan ssh is disabled

RickSommerville:
Merged bab7b187, thanks.

PR #309, opened 2021-06-11
Wifi-2635. Correcting the Model name format

netexp-ravi:
When the manufacturer block does not contain the model info,
it is then extracted from /tmp/sysinfo/board_name, stripped
off the manufacturer name and converted for all upper case string.


RickSommerville (approved):

RickSommerville:
Merged 463f9039, thanks.

PR #312, opened 2021-06-12
wifi-2512: apc gets stuck in wt mode 1 in 5 reboots

chaitanyagodavarthi:
Check if APC conf is disabled and not update the
APC state.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -821,6 +821,29 @@ struct schema_APC_Config apc_conf;
 void apc_state_set(struct blob_attr *msg)
 {
 	struct blob_attr *tb[__APC_ATTR_MAX] = { };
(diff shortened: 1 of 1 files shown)


RickSommerville (commented):
The description says if APC is disabled, then don't update the state.  However, you actually update the state to disabled.  This seems to be implementing all the parameters to disabled.  I can't tell if this will fix the problem.

Do you have a drawing of the state machine for this?  You have a small set of states and events.  I think you can draw it and see whatever race condition is causing issues.

chaitanyagodavarthi:
> The description says if APC is disabled, then don't update the state. However, you actually update the state to disabled. This seems to be implementing all the parameters to disabled. I can't tell if this will fix the problem.
> Do you have a drawing of the state machine for this? You have a small set of states and events. I think you can draw it and see whatever race condition is causing issues.

I have attached an image, for this race condition.
image>

RickSommerville:
OK, I understand the hole being plugged.  Why is the ubus notification WT/true occurring?

RickSommerville:
Merged b2a9e631, thanks

PR #313, opened 2021-06-14
Wifi-2538 : clients dropped on RRM rebalance

AmmadRehmatCUS:
- /sbin/relaoad_config file is modified to call another script in case the changes received are relevant to rrm only. In that case next script will do a channel switch using ubus


feeds/wlan-ap/opensync/files/bin/wlan_ap_rrm_rebalance.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+filename="/tmp/wireless_changes.txt"
+freq_table="1:2412 2:2417 3:2422 4:2427 5:2432 6:2437 7:2442 8:2447 9:2452 10:2457 11:2462 12:2467 13:2472 14:2484 36:5180 40:5200 44:5220 48:5240 52:5260 56:5280 60:5300 64:5320 100:5500 104:5520 108:5540 112:5560 116:5580 120:5600 124:5620 128:5640 132:5660 136:5680 140:5700 144:5720 149:5745 153:5765 157:5785 161:5805 165:5825"
+while IFS= read -r line; do
+        if [ "$line" = "phy0" ]; then
```
RickSommerville, line 5: There must be a more elegant way to do this.  Hard-coding the number of radios is not great.

feeds/wlan-ap/opensync/files/bin/wlan_ap_rrm_rebalance.sh
```diff
@@ -0,0 +1,29 @@
...
+done< "$filename"
+
+channel=$(uci get wireless.${radio}.channel)
+echo $channel
+for freq in $freq_table ; do
+        chan="${freq%%:*}"
+        frequency="${freq##*:}"
+        jstr='{"freq":'"$frequency"'}'
+        if [ "${channel}" = "${chan}" ]; then
+                ubus call hostapd.${interface} switch_chan "$jstr"
```
RickSommerville, line 27: In rrm_ubus.c/ubus_set_channel_switch(), Nagendrababu found it necessary to include additional parameters to get this to work (especially for WIFI6).  In particular, I don't see the "bcn_count", the "bandwidth" or the "sec_channel_offset". 

Did you test on a WIFI6 as well as WIFI5?
Did you actual sniff the air for the CSA in the beacon?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -371,18 +371,24 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
 
 	char phy[6];
 	char ifname[8];
+	bool CSA_reuested = false;
```
RickSommerville, line 374: CSA_requested.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -485,6 +497,17 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
 	if (changed->custom_options)
 		radio_config_custom_opt_set(&b, &del, rconf);
 
+	if(CSA_reuested) {
```
RickSommerville, line 500: I'd gather up all the test and set operations on CSA_reuested and this clause into a single function and call it here.  It will be more readable and more suitable to adding the other checks for the RRM cell-sizing step that will come next.

patches/0055-Wifi-2538-RRM-rebalance-disconnects-client.patch
```diff
@@ -0,0 +1,35 @@
...
+ 	file=${config##*/}
+ 	uci show "${file##*/}" > /var/run/config.check/$file
+ done
++FILE=/tmp/wireless_changes.txt
+ MD5FILE=/var/run/config.md5
+ [ -f $MD5FILE ] && {
+-	for c in `md5sum -c $MD5FILE 2>/dev/null| grep FAILED | cut -d: -f1`; do
+-		ubus call service event "{ \"type\": \"config.change\", \"data\": { \"package\": \"$(basename $c)\" }}"
++        for c in `md5sum -c $MD5FILE 2>/dev/null| grep FAILED | cut -d: -f1`; do
++                if [ -f "$FILE" ];then
```
RickSommerville, line 25: This condition also needs to ensure that the change is to /etc/config/wireless.  Eg $(basename $c) needs to equal wireless.
Otherwise you are blocking every other service from updating (and calling your function too many times).

patches/0055-Wifi-2538-RRM-rebalance-disconnects-client.patch
```diff
@@ -0,0 +1,35 @@
...
++                        bash /bin/ubus_update.sh
```
RickSommerville, line 26: At my request, you changed the name to:  wlan_ap_rrm_rebalance.sh.
You need to use the new name here.

You should not need the "bash /bin/" pre-fix.

feeds/wlan-ap/opensync/files/bin/wlan_ap_rrm_rebalance.sh
```diff
@@ -0,0 +1,29 @@
...
+        fi
+done
```
RickSommerville, line 29: Should somebody be deleting the flag file after it is consumed?

RickSommerville (changes requested):
Multiple changes are required.

Also, the WIFI-2538 JIRA ticket does not contain any test-cases.  This will need independent tests of WIFI5 and WIFI6 due to hostapd variation.  You should sniff the air to confirm that the CSA is sent in the beacon, and to verify the AP actually switches channel.

ArifAlam (commented):
Why is wlan_ap_led.sh part of commit?

patches/0055-Wifi-2538-RRM-rebalance-disconnects-client.patch
AmmadRehmatCUS: i will still need to use prefix bash /bin/ as we will need to provide the path to file

RickSommerville: /bin should automatically be searched.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -371,18 +371,26 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
...
-	if (changed->enabled)
+	if (changed->enabled) {
 		blobmsg_add_u8(&b, "disabled", rconf->enabled ? 0 : 1);
+		is_csa_requested = false;
+	}
+
 
 	if (changed->tx_power) {
 		int max_tx_power;
+		is_csa_requested = false;
```
RickSommerville, line 392: This is incomplete.  If the the tx-power is set to greater than the max_power the AP will set it to max_power and a change will be reported continuously.  This change shouldn't block the channel rebalance operation.

A similar issue will occur with tx_chainmask, rx_chainmask and with ht_mode when set to Auto.

I think we are going to need my proposed change to track the Auto behaviour and shut down the change indication.  Chaitanya already included part of it in WIFI-2649 yesterday.

Will need to think about this.  I don't think this code needs to change - rather additional code needs to prevent these change flags from being set.

RickSommerville (commented):
Need to solve the Auto problem.  Did your test cases cover profiles with Auto settings?

May be better for me to build onto this commit.

RickSommerville:
Obsolete.  New simpler solution found.

PR #314, opened 2021-06-14
wifi-2075- Fix for inconsistency in applying vif configuration

yashvardhan-netex:
- During the configuration process, AP was triggering network
   and wireless reload multiple times in a very short window
   resulting in a poorly configured hostapd. This patch makes sure
   that network/wireless is reloaded only once after all the configuration
   is committed to UCI files.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/command/src/node_config.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/command/src/node_config.c
@@ -159,7 +159,6 @@ static void syslog_handler(int type,
 	blob_to_uci_section(uci, "system", "@system[-1]", "system",
 			    b.head, &log_param, NULL);
 	uci_commit_all(uci);
-	system("/sbin/reload_config");
 	if (del)
 		node_state_del("syslog");
 	else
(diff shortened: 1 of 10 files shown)


RickSommerville (approved):
I like this, seems elegant solution.  Two questions:
1) I feel like the timer test capability you built must already exist somewhere.
2) Why on the OVSDB table delete side do you not update the timer?

RickSommerville (approved):
Looks good.  Two questions:
1.  I feel like timer functions equivalent to what you implemented must already exist.
2. Why not reset the counter on the delete paths?

yashvardhan-netex:
> Looks good. Two questions:
> 
> 1. I feel like timer functions equivalent to what you implemented must already exist.
> 2. Why not reset the counter on the delete paths?

1. We needed a countdown timer which we would reset whenever we receive a new configuration action(DEL/MOD/NEW). I looked into Uloop and libev before implementing this but none of them were solving our purpose. Uloop timer APIs works only with uloop_run which is a blocking call and when I tried to use it, it was blocking on one of the callbacks and not allowing others to execute unless the timer expires and we call Uloop_end. Yeah, But I took a good reference from these standard timers and adapted to suit our case.
2. I think I am resetting the counter on DEL operations. It is just that I am currently resetting it on Inet,RRM and hotspot table updates as VIF/Radio callback did not seem to work for some reason. But since all these profiles are pushed together and we have a timer expiry of 35 secs, we are good to go for now. I believe we can tackle the auto RRM exception later.

NagConnectus (approved):
I tried this load to check captive portal functionality, and it's working fine!

RickSommerville:
Merged 39bd8f30, thanks very much.

PR #315, opened 2021-06-15
WIFI-2634: Enable RADIUS accounting multi session id AVP

ArifAlam:
Turn on PMKSA and Opportunistic Key Caching to get Acct-Multi-Session-Id working.


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/vif.c
```diff
@@ -488,6 +490,7 @@ static int vif_config_security_set(struct blob_buf *b,
 				strcat(key_holder_str, key_str);
 				blobmsg_add_string(b, "r1kh", key_holder_str);
 			}
+			blobmsg_add_bool(b, "auth_cache", 1);
```
RickSommerville, line 493: Don't see anything deleting this for other setups.  Is that a potential problem?   Suspect not if we always tear-down before reconfigure.

RickSommerville (approved):

RickSommerville:
Merged 1b45173b, thanks

PR #319, opened 2021-06-17
WIFI-2152: Upgrade recovery

shimmy568:

Made is so that when an upgrade fails (with -9 error code) the temp files are deleted afterwards since that seemed to prevent us from attempting another upgrade.

Also patched opensync so we can attempt another upgrade without changing the timer felids and only changing the url field.

### DIFF
--- a/feeds/wlan-ap/opensync/files/bin/flash-firmware
+++ b/feeds/wlan-ap/opensync/files/bin/flash-firmware
@@ -40,6 +40,8 @@ uci commit
 /sbin/sysupgrade $IMGFILE
 if [ "$?" != "0" ] ; then
   echo "$0: Sysupgrade failed."
+  rm /tmp/sysupgrade.meta
+  rm /tmp/upgrade.*
   exit 1
 fi
 
--- a/feeds/wlan-ap/opensync/patches/39-allow-upgrade-retry.patch
+++ b/feeds/wlan-ap/opensync/patches/39-allow-upgrade-retry.patch
@@ -1,0 +1,12 @@
+--- a/src/um/src/um_ovsdb.c
++++ b/src/um/src/um_ovsdb.c
+@@ -356,7 +356,8 @@ static void callback_AWLAN_Node(
+                 //TODO Is there something that needs to be done here?
+             }
+ 
+-            if(awlan_node->upgrade_timer_changed){
++            if(awlan_node->upgrade_timer_changed
++                || ((awlan_node->firmware_url_changed) && (strlen(awlan_node->firmware_url) > 0))) {
+                 if (awlan_node->upgrade_timer > 0)
+                 {
+                     /* if there is active timer, stop it to set new value   */


yashvardhan-netex:
> I've been talking with Rahul and I believe the cloud will be implementing code to set the url to "" and then back to the URL that they want to retry with. With that assumption it should work

In that case, if cloud retries with the same url, AP would be downloading and validating the firmware again. Overkill? Is resetting upgrade_timer a better choice here?

shimmy568:
> > I've been talking with Rahul and I believe the cloud will be implementing code to set the url to "" and then back to the URL that they want to retry with. With that assumption it should work
> 
> In that case, if cloud retries with the same url, AP would be downloading and validating the firmware again. Overkill? Is resetting upgrade_timer a better choice here?

I don't believe that's an option since the 90 second value in hard coded in the opensync gateway. So I think our only option is doing this with the URL

rahulsharma2020:
> I've been talking with Rahul and I believe the cloud will be implementing code to set the url to "" and then back to the URL that they want to retry with. With that assumption it should work

Right. Cloud would be sending a _null_ URL in between 2 different retries, so the table column should clear. 
@shimmy568: Do you need an empty string instead of Null?

yashvardhan-netex:
> > > I've been talking with Rahul and I believe the cloud will be implementing code to set the url to "" and then back to the URL that they want to retry with. With that assumption it should work
> > 
> > 
> > In that case, if cloud retries with the same url, AP would be downloading and validating the firmware again. Overkill? Is resetting upgrade_timer a better choice here?
> 
> I don't believe that's an option since the 90 second value in hard coded in the opensync gateway. So I think our only option is doing this with the URL

But that hardcoding needs to go at some point as we do want to support scheduling of upgrade. Anyhow, we have make sure all the involved parties are in agreement before we proceed. @RickSommerville  Any inputs?

shimmy568:
> Right. Cloud would be sending a _null_ URL in between 2 different retries, so the table column should clear.
> @shimmy568: Do you need an empty string instead of Null?

Null is OK. As long as theres a change in the table the upgrade should trigger

feeds/wlan-ap/opensync/files/bin/flash-firmware
```diff
@@ -40,6 +40,8 @@ uci commit
 /sbin/sysupgrade $IMGFILE
 if [ "$?" != "0" ] ; then
   echo "$0: Sysupgrade failed."
+  rm /tmp/sysupgrade.meta
+  rm /tmp/upgrade.tar
```
yashvardhan-netex, line 44: I guess this could also be upgrade.bin for some targets?

shimmy568: Ah, good catch. I'll change this to upgrade.*

RickSommerville (approved):

RickSommerville:
Merged d6d45ba6, thanks.

PR #320, opened 2021-06-17
WIFI-2671: Added codes 12 & 13 for reboot & factory reset upgrade opt…

shimmy568:

We decided it would be easier for the AP to send up a different status code to the cloud instead of the cloud keeping track of what operation it started on the AP. 

A change will need to be made on the cloud side to trigger a different message box on the UI


RickSommerville (commented):
Can we rename the patch?   8)

shimmy568:
> Can we rename the patch? 8)

Whoops 🤦

RickSommerville:
Merged 208b0165, thanks

PR #321, opened 2021-06-18
Wifi-2538 : clients dropped on RRM rebalance

AmmadRehmatCUS:
- /sbin/relaoad_config file is modified to call another script in case the changes received are relevant to rrm only. In that case next script will do a channel switch using ubus


AmmadRehmatCUS:
TEST PR

PR #324, opened 2021-06-23
WIFI-2685, WIFI-2694: Fixed issues in Luci http portal

shimmy568:

Added a save button to the "Connectivity" and fixed the button on the LAN tab so it applied the correct settings

Moved a font that we were loading from the internet to the AP so it can load when not connected to the web.


feeds/wlan-ap/luci/luci-theme-tip/htdocs/luci-static/tip/cascade.css
```diff
@@ -1,4 +1,12 @@
-@import url('https://fonts.googleapis.com/css2?family=Montserrat&display=swap');
+/* latin */
```
shimmy568, line 1: Only added font's for "latin" since I assume we don't care about "cyrillic-ext", "cyrillic", "vietnamese" and "latin-ext" font sets. We could also change out the font for a standard web font to avoid this issue all together.

RickSommerville: OK

RickSommerville (approved):

PR #326, opened 2021-06-23
WIFI-2681: Allow re-trying a failed upgrade

shimmy568:

Instead of deleting the left over image from the /tmp/ directory after a failed call to `sysupgrade` I've got it now that `flash-firmware` can now operate with multiple files that match the pattern `upgrade.*` and only selects the most recent one. 

This doesn't solve the root issue but does fix the worst case brick scenario and also the issue where we are unable to re-try an upgrade once one fails. More testing should be done to determine why `sysupgrade` returns a non-zero code and then upgrades anyways, which is similar [NETEXP-1938](https://connectustechnologies.atlassian.net/browse/NETEXP-1938) and has been documented before.


RickSommerville (approved):

RickSommerville:
Merged 5bf882b0, thanks

PR #327, opened 2021-06-24
WIFI-2728-captive-portal-config-parameters-overwriting-issue

NagConnectus:
This patch will add the conditions to respond captive portal API's to
only valid configuration

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/captive.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/captive.c
@@ -651,17 +651,21 @@ void vif_captive_portal_set(const struct schema_Wifi_VIF_Config *vconf, char *if
 			blobmsg_add_string(&cap, "gatewayname", value);
 
 		else if (strcmp(opt, "splash_page_logo") == 0) {
-			blobmsg_add_string(&cap, "splash_page_logo", value);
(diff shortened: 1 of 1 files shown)


RickSommerville (commented):
So basically, update UCI and download the file when the name changes and is not NULL.

Question:  This means that the old file from when the filename value was non-NULL is still present and that the UCI still points to it.  Is that what you want?

Maybe we should be doing a blobmsg_add_string() to a delete blobmsg, and then using the delete parameter of blob_to_uci_section().   There are examples easily found.

NagConnectus:
> So basically, update UCI and download the file when the name changes and is not NULL.
> 
> Question: This means that the old file from when the filename value was non-NULL is still present and that the UCI still points to it. Is that what you want?
> 
> Maybe we should be doing a blobmsg_add_string() to a delete blobmsg, and then using the delete parameter of blob_to_uci_section(). There are examples easily found.
> 
> This comment may not apply if the old UCI section was deleted before the change.

Here my intention is when there are natted ssid's without captive portal profile on it, all the captive portal variables are null. there is no protection mechanism in such scenarios to not call curl api to download files. Since I have fixed file name on the AP for any given urls, this file is continuously getting downloading and deleting. To eliminate that issue I'm just calling curl api only when respective interface has captive portal configuration.

RickSommerville (approved):

RickSommerville:
Merged d852dcc8, thanks.

PR #328, opened 2021-06-24
WIFI-2825: Suppress noisy logs

shimmy568:

Removed the log that we have notifying us of the new value of conn_since.

I couldn't think of a practical way of just lowering the OVSDB Mon log for just this value so I lowered it to TRACE instead.


RickSommerville (approved):

RickSommerville:
Merged b6743c34, thanks

PR #331, opened 2021-06-29
WIFI-2691: Enable proxy arp by default

ArifAlam:
wifi6 AP does not forward broadcast ARP request sent from one wifi
client to another wifi client.  Enable proxy arp by default to have the
AP respond to ARP requests on behalf of the client.

### DIFF
--- a/feeds/wifi-ax/hostapd/files/hostapd.sh
+++ b/feeds/wifi-ax/hostapd/files/hostapd.sh
@@ -529,7 +529,7 @@ hostapd_set_bss_options() {
 	set_default signal_poll_time 5
 	set_default signal_drop_reason 3
 	set_default signal_strikes 3
-	set_default proxy_arp 0
+	set_default proxy_arp 1
 	set_default multicast_to_unicast 0
 
 


feeds/wifi-ax/hostapd/files/hostapd.sh
```diff
@@ -529,7 +529,7 @@ hostapd_set_bss_options() {
 	set_default signal_poll_time 5
 	set_default signal_drop_reason 3
 	set_default signal_strikes 3
-	set_default proxy_arp 0
+	set_default proxy_arp 1
```
RickSommerville, line 532: Should this behaviour be the same on WIFI5?

ArifAlam: wifi5 APs seem to forward broadcast ARP requests from one wireless client to another wireless client.  We do not need to enable proxy arp for this issue.  Enabling proxy arp in general by default as a feature for what it does is a good idea.  Perhaps can be done on the cloud side.

RickSommerville:
Merged 837b4511, thanks

PR #333, opened 2021-06-30
WIFI-2626: radius proxy: fix wildcard realm config

ArifAlam:
Wildcard realm needs to be configured as the very last item.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radius_proxy.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radius_proxy.c
@@ -361,6 +361,95 @@ static bool radius_proxy_config_set(struct schema_Radius_Proxy_Config *conf)
 	return true;
 }
 
+void radius_proxy_fixup()
+{
+	struct uci_package *radsecproxy;
+	struct uci_context *proxy_ctx;
+	struct uci_element *e = NULL, *tmp = NULL;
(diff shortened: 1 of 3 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radius_proxy.c
```diff
@@ -361,6 +361,95 @@ static bool radius_proxy_config_set(struct schema_Radius_Proxy_Config *conf)
 	return true;
 }
```
chaitanyagodavarthi, line 363: Is it not possible to check for "*" realm in radius_proxy_config_set itself in. "/* Configure realm block */" and add it at the end instead of deleting it and re-adding.

ArifAlam: The radius_proxy_config_set function is called per ovsdb row insert event in the Radius_Proxy_Config OVSDB table.  I can't tell if the ovsdb row with the wildcard is the last row when the radius_proxy_config_set function is called.

RickSommerville (commented):
Looks like it does what you want.  It seems clunky and very specifically tied to a full wildcard *.  Do we expect misbehaviour if other pets of the list are disordered?

I am trying to think of other ordered lists we might have.  SNTP servers maybe.

I wonder if this is better handled generally by Cloud supplied order.

Lets have a quick chat Monday.

RickSommerville:
Merged dedbb75a

PR #340, opened 2021-07-12
WIFI-2978 - Fix EC420 reset button

ArthurSu0211:
The EC420 reset button don't work, so it cannot be restarted or restored
to factory settings. The reset button is fixed to gpio18.

Signed-off-by: Arthur Su <arthur.su@tp-link.com>


RickSommerville:
Merged 95f43d6d, thanks.

PR #342, opened 2021-07-13
WIFI-2814: Removed duplicate entries from hostapd.sh for hs20_oper_friendly_name

shimmy568:
Simple fix, just had to remove the extra lines we had in each hostapd.sh file that was adding the duplicate entry to the config file


RickSommerville (approved):
Looks good.

RickSommerville:
Merged 52867a65, thanks

PR #344, opened 2021-07-13
Wifi-3021. Avoid creating VLAN interface on LAN side

netexp-ravi:
Under Bridged mode of configuration, when creating VLAN interface,
we create VLAN for both WAN and LAN side, whereas LAN side VLAN is
not requried.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/netifd/src/wifi_inet_config.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/netifd/src/wifi_inet_config.c
@@ -367,10 +367,24 @@ static void callback_Wifi_Inet_Config(ovsdb_update_monitor_t *mon,
 {
 	switch (mon->mon_type) {
 	case OVSDB_UPDATE_NEW:
+		if (iconf->parent_ifname_exists && iconf->vlan_id > 2) {
+			if(!strncmp(iconf->parent_ifname, "lan", strlen("lan"))) {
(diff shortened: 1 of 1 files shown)


RickSommerville (commented):
I expected to see some changes on the vif.c side.  Were they not required?

netexp-ravi:
> I expected to see some changes on the vif.c side. Were they not required?
> 
> Please put an explanation of change and of test cases in the JIRA ticket. Eg what is different in /etc/config/xyz.

Yes, no changes were required on vif.c side. 
I did update Wifi-3021 under description. Please let me know if more explanation is needed. Testing-wise, I just tested if the LAN side interface for VLAN is created or not. Checked on both Wifi6 and Wifi5 APs.

RickSommerville:
Did you actually connect a client and run traffic in each of the Wifi5 and Wifi6 cases?

shimmy568 (approved):
Tested on my EX227 and it seems to work as intended

netexp-ravi:
> Did you actually connect a client and run traffic in each of the Wifi5 and Wifi6 cases?

Done.
Seems OK with all 3 modes. Able to run traffic on client.
NAT mode. Client get DHCP IP from home router.
Bridge mode. No VLAN. Client get DHCP IP from home router.
Bridge mode. Custom VLAN + VLAN subnet DHCP Server on Mikrotik router. Client get DHCP IP from VLAN subnet.

RickSommerville:
Merged 17f86797, thanks

PR #347, opened 2021-07-16
WIFI-3063: Changed the way that we fetch the interface name so the in…

shimmy568:
I just changed the way that we get the name for the interface name from `ifstatus`. During the startup process the "l3_device" field isn't set but the "device" is. At least for our applications these fields have the same value once the interface is set up. 

Another possible fix to this issue is to add `sleep 10` to the top of the function that does the configuration setup. But that seems much more hacky so I decided to go with this fix


shimmy568:
Leaving this as a draft for now, still need to do some more testing.

NagConnectus (approved):
looks good to me!

ArifAlam:
Merged to pending. Thanks.

PR #348, opened 2021-07-18
ipq807x: fix the reset buttons on the CIG units

blogic:
Signed-off-by: John Crispin <john@phrozen.org>


ArifAlam:
Merged to pending. Thanks.

PR #351, opened 2021-07-23
wifi-2940: Add ath11k beacon and mcast rate

chaitanyagodavarthi:
Add support for changing mcast and beacon rate
in opensync for ath11k
Add debugfs entry for ath11k driver to set mcast,
beacon and mgmt rate.
For help:
cat /sys/kernel/debug/ieee80211/phyX/ath11k/set_rates

Eg: Set beacon rate to 12Mbps for iface wlan0 5G band
echo "wlan0 beacon 5 0x10000002" > /sys/kernel/debug/ieee80211/phy0/ath11k/set_rates

### DIFF
--- a/feeds/wifi-ax/mac80211/patches/pending/214-debugfs-set-rate-mgmt-mcast-bcn.patch
+++ b/feeds/wifi-ax/mac80211/patches/pending/214-debugfs-set-rate-mgmt-mcast-bcn.patch
@@ -1,0 +1,392 @@
+Index: backports-20200902_001-4.4.60-931c337125/drivers/net/wireless/ath/ath11k/debug.c
+===================================================================
+--- backports-20200902_001-4.4.60-931c337125.orig/drivers/net/wireless/ath/ath11k/debug.c
++++ backports-20200902_001-4.4.60-931c337125/drivers/net/wireless/ath/ath11k/debug.c
+@@ -1838,6 +1838,230 @@ static const struct file_operations fops
+ 	.open = simple_open
+ };
+ 
++static ssize_t ath11k_read_set_rates(struct file *file,
++				     char __user *user_buf,
++				     size_t count, loff_t *ppos)
++{
++	const char buf[] =
++		"This is to set fixed bcast, mcast, and beacon rates.  Normal rate-ctrl\n"
++		"is handled through normal API using 'iw', etc.\n"
++		"To set a value, you specify the dev-name, type, band and rate-code:\n"
++		"types: bcast, mcast, beacon, mgmt\n"
++		"bands: 2, 5\n"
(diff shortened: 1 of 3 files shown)


feeds/wifi-ax/mac80211/patches/pending/214-debugfs-set-rate-mgmt-mcast-bcn.patch
```diff
@@ -0,0 +1,392 @@
...
++	}
++	else if (set_rate_type == WMI_VDEV_PARAM_BCAST_DATA_RATE) {
++		arvif->bcast_rate[cfg_band] = rc;
++		arvif->bcast_rate_set[cfg_band] = true;
++	}
++	else if (set_rate_type == WMI_VDEV_PARAM_MCAST_DATA_RATE) {
++		arvif->mcast_rate[cfg_band] = rc;
++		arvif->mcast_rate_set[cfg_band] = true;
++	}
++	else if (set_rate_type == WMI_VDEV_PARAM_BEACON_RATE) {
```
ArifAlam, line 175: duplicate entry.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
```diff
@@ -200,14 +260,9 @@ int rrm_get_backup_channel(const char * freq_band)
 
 }
 
-bool rrm_noreload_config(struct schema_Wifi_RRM_Config *conf, const char *rname)
+bool rrm_noreload_config(struct schema_Wifi_RRM_Config *conf, const char *wlanif)
```
shimmy568, line 263: Not really part of this PR. But I think we can change this function to void, since we aren't actually returning anything else other than true 

(it looks like that if statement that this was called in before didn't even have anything in it in the first place)

shimmy568 (approved):

feeds/wifi-ax/mac80211/patches/pending/214-debugfs-set-rate-mgmt-mcast-bcn.patch
chaitanyagodavarthi: done

yashvardhan-netex (commented):
Tested on wifi5? Because there are some changes in the hotplug script.

chaitanyagodavarthi:
> Tested on wifi5? Because there are some changes in the hotplug script.

Yes, Tested on wifi5

yashvardhan-netex (approved):
Looks good!

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/rrm_config.c
chaitanyagodavarthi: Done

ArifAlam:
Merged to pending. Thanks.

PR #352, opened 2021-07-26
WIFI-3236: Channel hop event suppression when channel hop triggered from cloud

shimmy568:
Made it so we don't send a channel hop event to the cloud when the channel hop was triggered from the cloud side of things.

Didn't test the impact of switching channels due to interference.

The way I did this was by adding a "reason" param to the "switch_chan" ubus call. In this case a reason code of 0 means that the channel switch was triggered by high interfearence. In this case we do send the event to the cloud, but if we call the channel switch ubus call from rrm_radio_rebalance_channel we send a 1 as the reason code instead to avoid sending up the event.

### DIFF
--- a/feeds/wifi-ax/hostapd/patches/f00-019-channel-switch-event.patch
+++ b/feeds/wifi-ax/hostapd/patches/f00-019-channel-switch-event.patch
@@ -1,0 +1,64 @@
+--- a/src/ap/ubus.c
++++ b/src/ap/ubus.c
+@@ -531,6 +531,7 @@ enum {
+ 	CSA_VHT,
+ 	CSA_HE,
+ 	CSA_BLOCK_TX,
++	CSA_REASON,
+ 	__CSA_MAX
+ };
+ 
+@@ -545,6 +546,7 @@ static const struct blobmsg_policy csa_p
+ 	[CSA_VHT] = { "vht", BLOBMSG_TYPE_BOOL },
+ 	[CSA_HE] = { "he", BLOBMSG_TYPE_BOOL },
+ 	[CSA_BLOCK_TX] = { "block_tx", BLOBMSG_TYPE_BOOL },
++	[CSA_REASON] = { "reason", BLOBMSG_TYPE_INT32 },
+ };
+ 
+ #ifdef NEED_AP_MLME
+@@ -810,7 +812,7 @@ void hostapd_dfs_csa_timeout(void *eloop
+ 		   iface->conf->ieee80211ac, iface->conf->ieee80211ax);
+ 
+ 	ret = hostapd_enable_iface(iface);
+-	if (ret == 0)
++	if (ret == 0 && css->reason == 0)
+ 		hostapd_ubus_handle_channel_switch_event(iface,
+ 						HOSTAPD_UBUS_HIGH_INTERFERENCE,
+ 						iface->freq);
+@@ -994,6 +996,7 @@ hostapd_switch_chan(struct ubus_context
+ 	SET_CSA_SETTING(CSA_VHT, freq_params.vht_enabled, bool);
+ 	SET_CSA_SETTING(CSA_HE, freq_params.he_enabled, bool);
+ 	SET_CSA_SETTING(CSA_BLOCK_TX, block_tx, bool);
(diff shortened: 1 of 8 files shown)


feeds/wifi-ax/hostapd/patches/f00-019-channel-switch-event.patch
```diff
@@ -0,0 +1,55 @@
...
++	SET_CSA_SETTING(CSA_REASON, reason, u32);
+ 
+ 	wpa_printf(MSG_INFO, "%s: CSS freq=%d chan=%d sec_chan_off=%d, width=%d, seg0=%d, seg1=%d", __func__,
+ 		css.freq_params.freq,
+@@ -1015,7 +1018,9 @@ hostapd_switch_chan(struct ubus_context
+ 			return UBUS_STATUS_NOT_SUPPORTED;
+ 	}
+ 
+-	hostapd_ubus_handle_channel_switch_event(hapd->iface, HOSTAPD_UBUS_HIGH_INTERFERENCE, css.freq_params.freq);
++	if (css.reason == 0) {
```
chaitanyagodavarthi, line 32: there are couple of other places this hostapd_ubus_handle_channel_switch_event is called in the hostapd, I saw it in ubus.c and dfs.c . I think we need to add the reason check in the ubus.c where it called from hostapd_dfs_csa_timeout, since that's also triggered by rrm rebalance . please check if we need it for the one in the dfs.c.

feeds/wifi-trunk/hostapd/patches/905-channel-switch-event.patch
```diff
@@ -0,0 +1,55 @@
...
+ 
+ 
+ 	for (i = 0; i < hapd->iface->num_bss; i++) {
+@@ -793,7 +796,9 @@ hostapd_switch_chan(struct ubus_context
+ 			return UBUS_STATUS_NOT_SUPPORTED;
+ 	}
+ 
+-	hostapd_ubus_handle_channel_switch_event(hapd->iface,HOSTAPD_UBUS_HIGH_INTERFERENCE,  css.freq_params.freq);
++	if (css.reason == 0) {
++		hostapd_ubus_handle_channel_switch_event(hapd->iface, HOSTAPD_UBUS_HIGH_INTERFERENCE, css.freq_params.freq);
```
chaitanyagodavarthi, line 33: same comment as above applies here.

chaitanyagodavarthi (changes requested):

feeds/wifi-ax/hostapd/patches/f00-019-channel-switch-event.patch
shimmy568: The call in dfs.c I think can be left along since it seems to only trigger if we detect that radar started transmitting on a DFS channel, so this should only happen automatically and won't be triggered from the cloud. As for the other one I added the same `reason==0` since that seems to be the path we take for a regular channel switch on a DFS channel.

chaitanyagodavarthi (approved):
Looks good

feeds/wifi-ax/hostapd/patches/f00-019-channel-switch-event.patch
```diff
@@ -0,0 +1,64 @@
...
++	[CSA_REASON] = { "reason", BLOBMSG_TYPE_INT32 },
+ };
+ 
+ #ifdef NEED_AP_MLME
+@@ -810,7 +812,7 @@ void hostapd_dfs_csa_timeout(void *eloop
+ 		   iface->conf->ieee80211ac, iface->conf->ieee80211ax);
+ 
+ 	ret = hostapd_enable_iface(iface);
+-	if (ret == 0)
++	if (ret == 0 && css->reason == 0)
```
ArifAlam, line 24: Do we need css->reason == 0 check here?

shimmy568: I believe so. This function gets called from `hostapd_switch_chan` when we're trying to switch to a dfs channel. So we need to make sure it gets suppressed for that case too and not just the regular channel hops.

ArifAlam:
Merged to pending. Thanks.

PR #353, opened 2021-07-27
Wifi-3237: Fix missing and duplicate client events

yashvardhan-netex:
Some client events are duplicated and some of them are missing in the
client event report.

Root Cause 1: Processed clients events were not tagged in hostapd and
were thus getting duplicated during subsequent polling of client events.
Root Cause 2: Race between processing and clearing of client sessions
was leading to missing events.

Solution 1: Processed client events are tagged in hostapd and are reported only once.
Solution 2: Sessions are cleared only after they are processed by SM.

Also got rid of some unnecessary logs.

Acceptance Criteria: No duplicate client events in the mqtt report send by AP.
Sessions in the hostapd are successfully cleared after being processed by SM.

### DIFF
--- a/feeds/wifi-ax/hostapd/patches/f00-019-ubus-fix-client-events.patch
+++ b/feeds/wifi-ax/hostapd/patches/f00-019-ubus-fix-client-events.patch
@@ -1,0 +1,112 @@
+--- a/src/ap/ubus.c
++++ b/src/ap/ubus.c
+@@ -580,6 +580,7 @@ static int hostapd_clear_session(struct
+ 	{
+ 		if (session_id == rec->session_id) {
+ 			/* dec counter and delete */
++			wpa_printf(MSG_INFO, "%s Deleting client session with session id %llu", __func__, session_id);
+ 			cached_events_nr -= rec->rec_nr;
+ 			avl_delete(&hapd->ubus.rt_events, &rec->avl);
+ 			os_free(rec->records);
+@@ -633,6 +634,11 @@ static int hostapd_sessions(struct ubus_
+ 		/* messages for current session */
+ 		for (size_t i = 0; i < rec->rec_nr; i++) {
+ 			c_rec = &rec->records[i];
++			if (c_rec->processed) {
++				/* Record is already reported, continue */
++				continue;
++			}
++
+ 			/* check message type */
+ 			switch (c_rec->type) {
+ 			/* ClientAuthEvent */
+@@ -646,6 +652,7 @@ static int hostapd_sessions(struct ubus_
+ 				blobmsg_add_u32(&b_ev, "auth_status", p->auth_status);
+ 				blobmsg_add_string(&b_ev, "ssid", p->ssid);
+ 				blobmsg_close_table(&b_ev, t_msg);
++				c_rec->processed = true;
+ 				break;
+ 			}
+ 
+@@ -665,6 +672,7 @@ static int hostapd_sessions(struct ubus_
+ 				blobmsg_add_u8(&b_ev, "using11v", p->using11v);
+ 				blobmsg_add_string(&b_ev, "ssid", p->ssid);
+ 				blobmsg_close_table(&b_ev, t_msg);
++				c_rec->processed = true;
+ 				break;
+ 			}
+ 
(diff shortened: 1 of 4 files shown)


feeds/wifi-trunk/hostapd/patches/908-ubus-fix-client-events.patch
```diff
@@ -0,0 +1,111 @@
+--- a/src/ap/ubus.c
++++ b/src/ap/ubus.c
+@@ -578,6 +578,7 @@ static int hostapd_clear_session(struct
+ 	{
+ 		if (session_id == rec->session_id) {
+ 			/* dec counter and delete */
++			wpa_printf(MSG_INFO, "%s Deleting client session with session id %llu", __func__, session_id);
```
chaitanyagodavarthi, line 7: This can be DEBUG print I guess, many client going in and out will cause lot of prints.

feeds/wifi-ax/hostapd/patches/f00-019-ubus-fix-client-events.patch
```diff
@@ -0,0 +1,112 @@
+--- a/src/ap/ubus.c
++++ b/src/ap/ubus.c
+@@ -580,6 +580,7 @@ static int hostapd_clear_session(struct
+ 	{
+ 		if (session_id == rec->session_id) {
+ 			/* dec counter and delete */
++			wpa_printf(MSG_INFO, "%s Deleting client session with session id %llu", __func__, session_id);
```
chaitanyagodavarthi, line 7: This can be debug print I guess.

feeds/wifi-trunk/hostapd/patches/908-ubus-fix-client-events.patch
```diff
@@ -0,0 +1,111 @@
...
++				continue;
++			}
+ 			/* check message type */
+ 			switch (c_rec->type) {
+ 			/* ClientAuthEvent */
+@@ -644,6 +649,7 @@ static int hostapd_sessions(struct ubus_
+ 				blobmsg_add_u32(&b_ev, "auth_status", p->auth_status);
+ 				blobmsg_add_string(&b_ev, "ssid", p->ssid);
+ 				blobmsg_close_table(&b_ev, t_msg);
++				c_rec->processed = true;
```
chaitanyagodavarthi, line 26: Just a thought. why don't we just delete the records instead of marking them as processed.

yashvardhan-netex: These are individual events which are part of a client session and  client session as a whole is cleared by SM when it has processed all the events.

feeds/wifi-trunk/hostapd/patches/908-ubus-fix-client-events.patch
yashvardhan-netex: There is no easy way to turn hostapd debug at run time. It is an important log which tells if we were successfully able to clear session from hostapd after being processed by SM. I would like to keep it info atleast till the time we have resolved  other client events issues on the cloud side and have done testing with a greater number of clients. May be we can make it debug later?

feeds/wlan-ap/opensync/src/src/sm/src/ubus_collector.c
```diff
@@ -543,7 +537,6 @@ static void ubus_collector_hostapd_invoke(void *object_path)
 
 static void ubus_collector_complete_bss_cb(struct ubus_request *req, int ret)
 {
-	LOG(DEBUG, "ubus_collector_complete_bss_cb");
```
chaitanyagodavarthi, line 546: Why remove all debug messages? they might be useful for debugging

chaitanyagodavarthi (approved):
why remove the debug messages? apart from that Looks good

yashvardhan-netex:
> why remove the debug messages? apart from that Looks good

Those are unuseful logs just getting printed periodically during polling even when there is no action and thus making it difficult to focus on the important ones.

ArifAlam (approved):

ArifAlam:
Merged to pending. Thanks.

PR #354, opened 2021-07-27
netexp-2635:Fix Associated clients in case of roaming

chaitanyagodavarthi:
Associated clients list decrements when client roams
from vifX to vifY. In this case, the client is
first added to the global client list as part of it
associating with vifY (but in actual it doesnt add as
the entry for client already exists) and then deleted
from the same as part of dissassociating from vifX.

Fix is to change the vif the client entry points to
instead of add/delete the client entry.


yashvardhan-netex (approved):
Looks good!

yashvardhan-netex (commented):
Please update the commit messsage with TIP Jira defect number instead of Netexp

NagConnectus (approved):
looks good

ArifAlam:
Merged to pending. Thanks.

PR #355, opened 2021-07-27
Wifi-3246. Add chan_info data to existing survey info data

netexp-ravi:
The data from the chan_info events were overwriting the data
to the existing survey info counters. The data in the chan_info
events were Read-On-Clear where as the Survey info was maintained
as incremental stats and the application in the User space
performs the calculations based on incremental survey info stats.
As an example, the survey->time was always stored as 150 due to
chan_info event's Clear-On-Read nature. And this lead to the
difference calcuation as 0 at the User space application.
Solution is to add the incoming chan_info data to the existing
survey info stats, instead of replacing the existing data with new.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
@@ -275,6 +275,21 @@ bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t sca
 {
 	target_survey_record_t delta;
 
+	if (data_new->info.chan != data_old->info.chan) {
+		/* Do not consider the old channel's data */
+		memset(data_old, 0, sizeof(*data_old));
+	} else if ((data_new->duration_ms <= data_old->duration_ms) &&
+			(data_new->chan_busy <= data_old->chan_busy)) {
+		/*
+		 * Take care of survey data from CHAN_INFO events which are Read-On-Clear
+		 * in nature, whereas survey data from BSS_CHAN_INFO events are Read only.
+		 * Chan_info event down in the driver updates only the duration, noise floor
+		 * and the chan_busy data.
+		 */
+		data_new->duration_ms += data_old->duration_ms;
+		data_new->chan_busy += data_old->chan_busy;
+	}
+
 	LOGD("Survey convert scan_type %d chan %d duration_new:%llu duration_old:%llu "
 	     "busy_new %llu busy_old:%llu tx_new %llu tx_old:%llu rx_new %llu rx_old:%llu "
 	     "rx_self_new %llu rx_self_old:%llu",


blogic:
@netexp-ravi as I explaine din Slack this is the totally wrong approach

netexp-ravi:
> @netexp-ravi as I explaine din Slack this is the totally wrong approach

Thanks, John. 
I will move the changes to the user space.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
```diff
@@ -275,6 +275,21 @@ bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t sca
 {
 	target_survey_record_t delta;
 
+	if (data_new->info.chan != data_old->info.chan) {
+		/* Do not consider the old channel's data */
+		memset(data_old, 0, sizeof(*data_old));
+	} else if ((data_new->duration_ms <= data_old->duration_ms) &&
+			(data_new->chan_busy <= data_old->chan_busy)) {
```
RickSommerville, line 282: Do we need to worry about a false trigger of this code due to a counter wrap?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
```diff
@@ -275,6 +275,21 @@ bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t sca
...
+		/*
+		 * Take care of survey data from CHAN_INFO events which are Read-On-Clear
+		 * in nature, whereas survey data from BSS_CHAN_INFO events are Read only.
+		 * Chan_info event down in the driver updates only the duration, noise floor
```
RickSommerville, line 286: Why are we getting two different kinds of events?  If the Chan_info event only contains duration, noise floor and chan_busy, then all the other counters will be zero and will get reported as such.  This willl cause an awkward discontinuity in the graphs.

Since we are asking for and expecting the BSS_CHAN_INFO wouldn't be better to report that only?  Although, this might be tough if the other event is occurring and resetting counters.

RickSommerville (commented):
Questions only, no changes requested.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
```diff
@@ -275,6 +275,21 @@ bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t sca
 {
 	target_survey_record_t delta;
 
+	if (data_new->info.chan != data_old->info.chan) {
+		/* Do not consider the old channel's data */
+		memset(data_old, 0, sizeof(*data_old));
```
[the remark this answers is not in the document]
netexp-ravi: I think that is not true. In my debug logs, I saw data not reset to zero after a channel change. I will cross-check this again.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
netexp-ravi: It is highly unlikely to have both the counters (duration_ms and chan_busy) wrap at the same time.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
netexp-ravi: We still don't know why do we get the chan_info events. I do not see any code requesting for this event, and it is not that this event is always generated for the current operating channel. It may be that the firmware is auto-generating these events based on some criteria or a bug.

At the driver level (in wmi.c), we see some valid info in the other parameters of chan_info event. I mean, we do see the rx and rx_self info being populated by the firmware. But, the same is not propagated all the way to the user-space. We do not see much of a variation due to this, as the measurement time for chan_info always seemed to be 150ms. So, we are stretching the missing parameters like rx and rx_self for another 150ms using the same old values we read from bss_chan_info.

I still believe, the right thing to do here is to patch the driver to update all the parameters under chan_info event processing. But for some reason, John does not like that idea. I did not understand the reason for not liking it though.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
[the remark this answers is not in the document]
RickSommerville: So if we reset the data_old to 0, then the delta from data_new -> 0 could be very large and incorrect.
Perhaps better if we just report all the values (including the duration) as 0, so that this transitional sample has no impact on the measurement.  Then at the next sample we will have good values of data_new and data_old from the same channel.

Discussion thoughts only, I don't know if the surmised effect happens nor if the cloud side can handle a sample with all values including duration = 0.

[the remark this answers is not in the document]
netexp-ravi: Very large or incorrect may be possible with this change.
If we report any data with duration=0, either the GW or the UI see that as a missing report. We see a gap in UI during that report.

We could remove this channel-change related new code and live with differentiating between the old channel's data and the new.

RickSommerville (approved):

PR #357, opened 2021-07-28
WIFI-2844: Removed the fallback case for when country code isn't set

shimmy568:
If the country code was blank in `/etc/config/wireless` this if statement would make it look like it was set to "CA" in the ovsdb tables. This means that when the cloud set it to "CA" we wouldn't update the `/etc/config/wireless` table since according to opensync it was already set to "CA". (this also meant this error would only happen if the region code was set to CA)

This fix seems to avoid the worst case scenario but the when the AP is booted with this fix (and having the country code entries removed from /etc/config/wireless) it seems to be put into backoff on it's first attempt to connect to the cloud but after that it's able to connect successfully on it's own after a little while.


shimmy568:
Seems like setting the country code to "00" sends the client into permanent backoff, so unless we get a fix done on the cloud side of things that would allow for that this is the best solution for now

ArifAlam:
Merged to pending. Thanks.

PR #360, opened 2021-07-30
WIFI-3306: Updated the LLDP description with firmware version info an…

shimmy568:
Updated the LLDPD configuration so that we use our firmware version in the description instead of the openwrt version.

Also I updated added rules so that the br-lan IP addresses aren't selected instead of the br-wan IP address since LLDPD just uses the first IP address in comes across in ifconfig. Tried with just PKG_RELEASE=1 but we need PKG_RELEASE=2 for the management pattern setting to work.

Used [this](https://github.com/openwrt/openwrt/commit/128250e8fc615aebe05642f023375dd7902808e1) openwrt commit as a reference for the wan IP change

### DIFF
--- a/patches/0058-LLDP-firmware-version.patch
+++ b/patches/0058-LLDP-firmware-version.patch
@@ -1,0 +1,68 @@
+From 5e5917308f769f1228c572e4aed4f8f7f6434a14 Mon Sep 17 00:00:00 2001
+From: Owen Anderson <owenthomasanderson@gmail.com>
+Date: Wed, 4 Aug 2021 13:38:16 -0400
+Subject: [PATCH] Updated LLDPD version to 1.0.5 and updated config information
+
+---
+ package/network/services/lldpd/Makefile         |  4 ++--
+ package/network/services/lldpd/files/lldpd.init | 11 ++++++-----
+ 2 files changed, 8 insertions(+), 7 deletions(-)
+
+diff --git a/package/network/services/lldpd/Makefile b/package/network/services/lldpd/Makefile
+index 5b36155b71..48a3993492 100644
+--- a/package/network/services/lldpd/Makefile
++++ b/package/network/services/lldpd/Makefile
+@@ -8,12 +8,12 @@
+ include $(TOPDIR)/rules.mk
+ 
+ PKG_NAME:=lldpd
+-PKG_VERSION:=1.0.3
++PKG_VERSION:=1.0.5
+ PKG_RELEASE:=1
+ 
+ PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


patches/0058-LLDP-firmware-version.patch
```diff
@@ -0,0 +1,66 @@
...
+ 
+ find_release_info()
+ {
+-	[ -s /etc/os-release ] && . /etc/os-release
+-	[ -z "$PRETTY_NAME" ] && [ -s /etc/openwrt_version ] && \
+-		PRETTY_NAME="$(cat /etc/openwrt_version)"
+-
+-	echo "${PRETTY_NAME:-Unknown OpenWrt release} @ $(cat /proc/sys/kernel/hostname)"
++	FW_VERSION=`sed -n 's/FW_IMAGE_ACTIVE:\(.*\)/\1/p' < /usr/opensync/.versions`
++	echo "${FW_VERSION:-Unknown release} @ $(cat /proc/sys/kernel/hostname)"
```
RickSommerville, line 40: Might be nice to specify "Unknown TIP OpenAp release" here.

patches/0058-LLDP-firmware-version.patch
```diff
@@ -0,0 +1,66 @@
...
+ }
+ 
+ write_lldpd_conf()
+@@ -48,6 +45,10 @@ write_lldpd_conf()
+ 	[ -n "$lldp_description" ] && echo "configure system description" "\"$lldp_description\"" >> "$LLDPD_CONF"
+ 	[ -n "$lldp_hostname" ] && echo "configure system hostname" "\"$lldp_hostname\"" >> "$LLDPD_CONF"
+ 
++	# Prevent br-lan IP from being used for LLDP
++	echo "configure system ip management pattern !192.168.1.1" >> "$LLDPD_CONF"
++	echo "configure system ip management pattern !192.168.0.1" >> "$LLDPD_CONF"
```
RickSommerville, line 50: Not sure what this is doing.  There is some possibility that if Lan is 192.168.0.1 then the Wan might be 192.168.1.1 or vice versa.  Let's discuss.

RickSommerville: Should be able to specify an interface eg br-wan or wan.  Per https://manpages.debian.org/testing/lldpd/lldpcli.8.en.html

RickSommerville (changes requested):

patches/0058-LLDP-firmware-version.patch
shimmy568: Did that, had to update lldpd to 1.0.5 for that feature to work though https://github.com/lldpd/lldpd/releases/tag/1.0.5

RickSommerville (approved):

RickSommerville:
Merged 80bfd398, thanks

PR #361, opened 2021-07-30
Wifi-3309: Fix Tx power reporting

yashvardhan-netex:
- AP was not reporting the Channel Tx power currently set in
   the firmware. This patch reads the Tx power from debugFS stats
   and reports it to cloud through Wifi_Radio_State table

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
@@ -216,6 +216,76 @@ static void update_channel_max_power(char* phy, struct schema_Wifi_Radio_State *
 	}
 }
 
+static int get_channel_tx_power_debugfs(char *stats_path, int *tx_power)
+{
+	bool found = false;
+	char str[512];
+	FILE * fp = NULL;
+	int txpower = -1;
+	int ret = -1;
+
+	fp = fopen(stats_path, "r");
+	if (!fp) {
+		LOGE("%s: Failed to get tx power, stats file does not exist", __func__);
+		return -1;
+	}
+
+	const char *substr = "Channel TX power";
+
+	while(fgets(str, 512, fp)) {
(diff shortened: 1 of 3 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -216,6 +216,76 @@ static void update_channel_max_power(char* phy, struct schema_Wifi_Radio_State *
...
+static int get_channel_tx_power(char* phy, int *tx_power)
+{
+	char ath_driver[ATH_DRIVER_NAME_LEN] = {'\0'};
+	char stats_path[128];
+
+	if (phy_get_ath_driver_name(phy, ath_driver)) {
+		LOGE("%s: Failed to get ATH driver name", __func__);
+		return -1;
+	}
+	if (!strcmp(ath_driver, "ath10k")) {
```
chaitanyagodavarthi, line 268: Use strncmp instead of strcmp in all places.

yashvardhan-netex: Both strings passed are null terminated. We should be fine. See ath_driver initialization above.

chaitanyagodavarthi: Good practice it is anyway. if somehow in the below statement whole length is filled with garbage, the '\0' will also get erased.
strncpy(ath_driver, basename(gl.gl_pathv[0]), ATH_DRIVER_NAME_LEN);

yashvardhan-netex: Agree, it is a good practice :). strncpy does not add null to the destination string. ATH_DRIVER_NAME_LEN-1 would be alright if the destination is already null terminated which it is in our case. I will make the necessary changes.

ArifAlam (approved):
Looks good.

PR #364, opened 2021-08-04
wifi-3051: Set proper mode & freq in RRM NF chan switch

chaitanyagodavarthi:
For RRM noise floor driven channel change:
1. Set HT VHT HE based on the hw_mode
2. Set secondary frequency (center_freq1) based on bandwidth


NagConnectus (approved):
Looks good!, tried staging load, and tested the functionality by injecting noise on the primary channel. Whenever noise floor value exceeds the threshold value on primary channel, the channel switching event is triggering and channel is switching to backup channel as expected and also staying on same hardware mode.

RickSommerville:
Merged  5ef05564, thanks.

PR #365, opened 2021-08-09
wifi-3258: Fix bridge client traffic on same wifi interface

chaitanyagodavarthi:
Allow bridging broadcast and multicast traffic within same
bridge port if hairpinmode is enabled for that port.
This fixes the ARP packet between clients on same wifi interface
being dropped.


NagConnectus (approved):
looks good!

ArifAlam (approved):

PR #366, opened 2021-08-13
Wifi-3463. Fix for scan timeout with scan on same interface

netexp-ravi:
Successive scan requests on same interface cause scan timeout.
Scan requests are enqueued at the Opensync layer and is designed
to handle the requests sequentially. If there are successive
scan requests for the same interface, then we end up deleteling
an on-going scan context. This is due to the interface name being
used as key for the AVL entries, and we were trying to reuse the
AVL entry if one already exist.
Solution is to always allocate a new entry without any reuse.


yashvardhan-netex (approved):
Looks good!

RickSommerville:
Merged to pending 4beda3ab, thanks

PR #367, opened 2021-08-13
WIFI-3244-WF610D-Config-Reset-Over-Sysupgrade

NagConnectus:
This patch will fix configuration restore problem over sysupgrade
on WF610D
Even though there is a configuration data in the /dev/mtdblock12, it is getting erased while reading it beacuse of JFFS2 empty bitmask code. So I added a condition to check whether the datatype available is JFFS2 or not, and making sure it is not empty before returning the JFFS2 flag to restore the configuration.

### DIFF
--- a/patches/0060-wifi-3244-wf610d-config-reset-over-sysupgrade.patch
+++ b/patches/0060-wifi-3244-wf610d-config-reset-over-sysupgrade.patch
@@ -1,0 +1,37 @@
+From 87af8ae014551b73a51c30a0c30da2a2c666e548 Mon Sep 17 00:00:00 2001
+From: Nagendrababu <nagendrababu.bonkuri@connectus.ai>
+Date: Fri, 13 Aug 2021 11:41:25 -0400
+Subject: [PATCH] WIFI-3244-config-reset-over-sysupgrade
+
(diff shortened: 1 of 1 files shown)


patches/0060-wifi-3244-wf610d-config-reset-over-sysupgrade.patch
```diff
@@ -0,0 +1,35 @@
+From 297d499d7ff5837e3f76df6edb098aaee2e33a54 Mon Sep 17 00:00:00 2001
+From: Nagendrababu <nagendrababu.bonkuri@connectus.ai>
+Date: Fri, 13 Aug 2021 11:41:25 -0400
+Subject: [PATCH] WIFI-3244-config-reset-over-sysupgrade
+
+This patch will add the support to restore the configuration over sysupgrade for wf610d AP
```
chaitanyagodavarthi, line 6: Here it doesn't explain the patch

chaitanyagodavarthi (approved):

patches/0060-wifi-3244-wf610d-config-reset-over-sysupgrade.patch
NagConnectus: done

PR #368, opened 2021-08-16
wifi-ax: Fix build of batman-adv against QCA 11.4 CS' mac80211

ecsv:
@blogic 

The new mac80211 version introduced various new compat code definitions which are now conflicting with the one in the compat-qca.h

Just dropping them is enough to get it compiling again.


blogic:
swesome, Thanks !

PR #369, opened 2021-08-17
Wifi-3490. Opensync workaround for 0 NF reported.

netexp-ravi:
Sometimes, the driver reports 0 Noise Floor value which is
not a valid data point. Adding a workaround in Opensync layer
to consider previous recorded NF value when 0 NF is reported.


yashvardhan-netex (approved):
Looks good!

ArifAlam (approved):

PR #371, opened 2021-08-24
Wifi-3522. Correcting an earlier commit for scan failures.

netexp-ravi:
Correcting an earlier commit 4beda3ab6607825639f9c018456a83a8e79f2d72
to consider all the cases where the avl entry needs be deleted when
a scan is either completed or aborted for any reason.

### DIFF
--- a/feeds/wlan-ap/opensync/patches/42-sm_dbg_log.patch
+++ b/feeds/wlan-ap/opensync/patches/42-sm_dbg_log.patch
@@ -2,20 +2,22 @@ Index: opensync-2.0.5.0/src/sm/src/sm_scan_schedule.c
 ===================================================================
 --- opensync-2.0.5.0.orig/src/sm/src/sm_scan_schedule.c
 +++ opensync-2.0.5.0/src/sm/src/sm_scan_schedule.c
-@@ -155,6 +155,12 @@ clean:
+@@ -155,6 +155,14 @@ clean:
  
      /* Remove processed context */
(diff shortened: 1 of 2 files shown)


NagConnectus (approved):
Looks good to me!

feeds/wlan-ap/opensync/patches/43-sm_del_avl_when_remove_scan_ctx.patch
```diff
@@ -0,0 +1,20 @@
...
+ #include "sm.h"
++#include "nl80211.h"
+ 
+ #define MODULE_ID LOG_MODULE_ID_MAIN
+ 
+@@ -76,6 +77,7 @@ void sm_scan_ctx_free(sm_scan_ctx_t *sca
+ {
+     if (NULL != scan_ctx)
+     {
++        nl80211_scan_remove(scan_ctx->scan_request.radio_cfg->if_name, scan_ctx);
```
yashvardhan-netex, line 17: As discussed, please delete the nl80211 avl entry in stats_nl80211.c itself. This way we can still keep the stats_nl80211 abstracted from this upper layer.

yashvardhan-netex (commented):
Please also do the other minor improvements as discussed.

yashvardhan-netex (approved):
Looks good! This does not solve the SM crash issue but we can go ahead with the merge if in anyway the change is not aggravating the crash issue.

PR #373, opened 2021-08-28
wifi-3613: Fix channel fallback after switch

yashvardhan-netex:
Radio was falling back to original channel after switch because
of a mismatch between radio state and radio config.

Updated the radio config with “switch-to” channel to match with the radio state.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
@@ -15,6 +15,7 @@ struct uci_context *uci;
 int reload_config = 0;
 static struct ev_timer  rrm_nf_timer;
 static void rrm_nf_timer_handler(struct ev_loop *loop, ev_timer *timer, int revents);
+ovsdb_table_t table_Wifi_Radio_Config;
 
 enum {
 	WIF_ATTR_PROBE_ACCEPT_RATE,
@@ -163,6 +164,8 @@ void rrm_nf_timer_handler(struct ev_loop *loop, ev_timer *timer, int revents)
 	uint32_t                 noise;
 	int32_t                  nf;
 	rrm_config_t             *rrm_config;
+	struct schema_Wifi_Radio_Config config = {};
+	json_t *iter;
 
 	ds_tree_t *radio_list = rrm_get_radio_list();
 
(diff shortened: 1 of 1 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/rrm/src/rrm_channel.c
```diff
@@ -224,9 +227,20 @@ void rrm_nf_timer_handler(struct ev_loop *loop, ev_timer *timer, int revents)
...
+			/*
+			 * Update the Wifi_Radio_Config table with the "switch-to" channel.
+			 * This is done to avoid the mismatch between radio state and radio config
+			 * because a mismatch would cause the radio to fallback to the original channel.
+			 */
+			iter = ovsdb_where_simple(SCHEMA_COLUMN(Wifi_Radio_Config, if_name), radio->schema.if_name);
+			if (iter) {
+				if (ovsdb_table_select_one_where(&table_Wifi_Radio_Config, iter, &config))
+					config.channel = rrm_config->rrm_data.backup_channel;
+			}
```
chaitanyagodavarthi, line 239: do we still need to do ubus call for channel switch, since if you change the radio_config it would change the channel.

chaitanyagodavarthi:
> Here changing radio config would in fact avoid the channel change from radio.c as now radio config and state would be same after the switch. Tested and working fine.

I think there is a race between this code and the radio.c radio config code, because between changing radio_config channel and the actual channel change (picked up by radio_state), the radio.c radio config code my also try to change the channel, but not sure how that double channel change will impact.

yashvardhan-netex:
> > Here changing radio config would in fact avoid the channel change from radio.c as now radio config and state would be same after the switch. Tested and working fine.
> 
> I think there is a race between this code and the radio.c radio config code, because between changing radio_config channel and the actual channel change (picked up by radio_state), the radio.c radio config code my also try to change the channel, but not sure how that double channel change will impact.

A mismatch between radio and state is checked at a periodic interval and even if it comes into picture, we are writing the same channel in radio config, so do not see much of an impact. However, we are not hitting the scenario in our testing. Puneet also confirmed and the channel is switching only once. IMO, not ideal but is probably the best solution before we move to ucentral.

chaitanyagodavarthi:
> > > Here changing radio config would in fact avoid the channel change from radio.c as now radio config and state would be same after the switch. Tested and working fine.
> > 
> > 
> > I think there is a race between this code and the radio.c radio config code, because between changing radio_config channel and the actual channel change (picked up by radio_state), the radio.c radio config code my also try to change the channel, but not sure how that double channel change will impact.
> 
> A mismatch between radio and state is checked at a periodic interval and even if it comes into picture, we are writing the same "switch-to" channel in radio config, so do not see much of an impact. However, we are not hitting the scenario in our testing. Puneet also confirmed and the channel is switching only once. IMO, not ideal but is probably the best solution before we move to ucentral.

Ok, Is any DFS channel tested?

yashvardhan-netex:
> > > > Here changing radio config would in fact avoid the channel change from radio.c as now radio config and state would be same after the switch. Tested and working fine.
> > > 
> > > 
> > > I think there is a race between this code and the radio.c radio config code, because between changing radio_config channel and the actual channel change (picked up by radio_state), the radio.c radio config code my also try to change the channel, but not sure how that double channel change will impact.
> > 
> > 
> > A mismatch between radio and state is checked at a periodic interval and even if it comes into picture, we are writing the same "switch-to" channel in radio config, so do not see much of an impact. However, we are not hitting the scenario in our testing. Puneet also confirmed and the channel is switching only once. IMO, not ideal but is probably the best solution before we move to ucentral.
> 
> Ok, Is any DFS channel tested?

Not needed. No change in the DFS code leg. We do not configure DFS channels as backup channel

chaitanyagodavarthi (approved):

PR #375, opened 2021-09-02
NETEXP-2839: Fix apc / radsec config

ArifAlam:
Update config on APC topology change.


yashvardhan-netex (commented):
How often this process is started/stopped? There are places where we are holding reload_config even after writing to UCI. For example to avoid clients disconection in case of RRM channel rebalancing. We will have to make sure this change does not impact such scenarios.

jaspreetsachdev:
@ArifAlam please use only TIP OpenWiFi JIRA references here.

PR #377, opened 2021-09-10
WIFI-3905-WIFI5-AP does not forward dhcp broadcast offer

NagConnectus:
This patch will add fix to allow DHCP broadcast packets, so that AP can forward DHCP broadcast offer to clients
Cause:
 In the bridge source code, there is a mechanism to drop all the broadcast packets if the proxy arp is enabled, because of this dhcp clients which set broadcast flag to 1 while connecting toAP's interfaces are not able get an ip address hence they are not connecting
 Solution:
Added the code to the bridge source code so that before dropping broadcast packets it first checks, whether it is dhcp broadcast packet or not. So that it won't drop dhcp braodcast packets.Hence DORA process will succeed and clients which sets broadcast flag to 1 gets an ip address even though the proxy arp is enabled.

### DIFF
--- a/patches/0063-Allow-DHCP-broadcast-packets.patch
+++ b/patches/0063-Allow-DHCP-broadcast-packets.patch
@@ -1,0 +1,84 @@
+From 2d71809001b7d05afbc6ebd98878826dda135659 Mon Sep 17 00:00:00 2001
+From: Nagendrababu <nagendrababu.bonkuri@connectus.ai>
+Date: Fri, 10 Sep 2021 12:47:12 -0400
+Subject: [PATCH] WIFI-3905-AP does not forward dhcp broadcast offer
+
+This patch will add fix to allow DHCP broadcast packets, so that AP can forward DHCP broadcast offer to clients
+
+cause:
+In the bridge source code, there is a mechanism to drop all the broadcast packets if the proxy arp is enabled, because of this dhcp clients which set broadcast flag to 1 while connecting toAP's interfaces are not able get an ip address hence they are not connecting
+solution:
+Added the code to the bridge source code so that before dropping broadcast packets it first checks, whether it is dhcp broadcast packet or not. So that it won't drop dhcp braodcast packets.Hence DORA process will succeed and clients which sets broadcast flag to 1 gets an ip address even though the proxy arp is enabled.
+---
+ .../A00-Allow-DHCP-braodcast-packet.patch     | 59 +++++++++++++++++++
+ 1 file changed, 59 insertions(+)
+ create mode 100644 target/linux/ipq40xx/patches-4.14/A00-Allow-DHCP-braodcast-packet.patch
+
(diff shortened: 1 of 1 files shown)


patches/0063-Allow-DHCP-broadcast-packets.patch
```diff
@@ -0,0 +1,71 @@
...
++--- linux-4.14.187.orig/net/bridge/br_forward.c
+++++ linux-4.14.187/net/bridge/br_forward.c
++@@ -20,6 +20,37 @@
++ #include <linux/if_vlan.h>
++ #include <linux/netfilter_bridge.h>
++ #include "br_private.h"
+++#include <linux/ip.h>
+++#include <linux/udp.h>
+++
+++static struct bootp_pkt {	/* BOOTP packet format */
```
chaitanyagodavarthi, line 29: is there a way not to add this whole structure here and detect the dhcp packets?

chaitanyagodavarthi (commented):
Please test it for Vlan case.

chaitanyagodavarthi (commented):
Add more in the commit message on why we are adding this fix, what's the cause and solution.

NagConnectus:
> Please test it for Vlan case.

Yes, tested with VLAN setup. It works fine

patches/0063-Allow-DHCP-broadcast-packets.patch
NagConnectus: To detect the DHCP packet there is local structure in the source code. And didn't find any global structure definitions to detect the DHCP packet.

NagConnectus:
> Add more in the commit message on why we are adding this fix, what's the cause and solution.

sure

patches/0063-Allow-DHCP-broadcast-packets.patch
```diff
@@ -0,0 +1,71 @@
...
+++}
++ 
++ /* Don't forward packets to originating port or forwarding disabled */
++ static inline int should_deliver(const struct net_bridge_port *p,
++@@ -205,7 +236,7 @@ void br_flood(struct net_bridge *br, str
++ 		if (p->flags & BR_PROXYARP)
++ 			continue;
++ 		if ((p->flags & BR_PROXYARP_WIFI) &&
++-		    BR_INPUT_SKB_CB(skb)->proxyarp_replied)
+++		    BR_INPUT_SKB_CB(skb)->proxyarp_replied && !dhcp_packet_check(skb))
```
chaitanyagodavarthi, line 65: put a comment here on why this is needed.

NagConnectus:
> Add more in the commit message on why we are adding this fix, what's the cause and solution.

done

patches/0063-Allow-DHCP-broadcast-packets.patch
NagConnectus: done

chaitanyagodavarthi (approved):

ArifAlam (approved):

PR #378, opened 2021-09-10
WIFI-3905-WIFI6-AP does not forward dhcp broadcast offer

NagConnectus:
This patch will add fix to allow DHCP broadcast packets,
so that WIFI6 AP can forward DHCP broadcast offer to clients
Cause:
In the bridge source code, there is a mechanism to drop all the broadcast packets if the proxy arp is enabled, because of this dhcp clients which set broadcast flag to 1 while connecting toAP's interfaces are not able get an ip address hence they are not connecting
Solution:
Added the code to the bridge source code so that before dropping broadcast packets it first checks, whether it is dhcp broadcast packet or not. So that it won't drop dhcp braodcast packets.Hence DORA process will succeed and clients which sets broadcast flag to 1 gets an ip address even though the proxy arp is enabled.

### DIFF
--- a/patches/0062-Allow-DHCP-Broadcast-packets.patch
+++ b/patches/0062-Allow-DHCP-Broadcast-packets.patch
@@ -1,0 +1,86 @@
+From c914392be2dbae30ceb139ed871e57c776c66f64 Mon Sep 17 00:00:00 2001
+From: Nagendrababu <nagendrababu.bonkuri@connectus.ai>
+Date: Thu, 9 Sep 2021 17:06:08 -0400
+Subject: [PATCH] WIFI-3905-AP does not forward dhcp broadcast offer
+
+This patch will add fix to allow DHCP broadcast packets, so that AP can
+forward DHCP broadcast offer to clients
+
+Cause:
+In the bridge source code, there is a mechanism to drop all the broadcast packets if the proxy arp is enabled, because of this dhcp clients which set broadcast flag to 1 while connecting toAP's interfaces are not able get an ip address hence they are not connecting
+Solution:
+Added the code to the bridge source code so that before dropping broadcast packets it first checks, whether it is dhcp broadcast packet or not. So that it won't drop dhcp braodcast packets.Hence DORA process will succeed and clients which sets broadcast flag to 1 gets an ip address even though the proxy arp is enabled.
+
+Signed-off-by: Nagendrababu <nagendrababu.bonkuri@connectus.ai>
+---
+ .../113-Allow-DHCP-broadcast-packet.patch     | 58 +++++++++++++++++++
+ 1 file changed, 58 insertions(+)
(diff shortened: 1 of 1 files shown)


patches/0062-Allow-DHCP-Broadcast-packets.patch
```diff
@@ -0,0 +1,74 @@
...
++--- linux-4.4.60-qsdk-10fd7d14853b7020b804acae690c8acec5d954ce.orig/net/bridge/br_forward.c
+++++ linux-4.4.60-qsdk-10fd7d14853b7020b804acae690c8acec5d954ce/net/bridge/br_forward.c
++@@ -20,6 +20,37 @@
++ #include <linux/if_vlan.h>
++ #include <linux/netfilter_bridge.h>
++ #include "br_private.h"
+++#include <linux/ip.h>
+++#include <linux/udp.h>
+++
+++struct bootp_pkt {		/* BOOTP packet format */
```
chaitanyagodavarthi, line 32: is there way to detect the dhcp packet without copying this whole structure here?

patches/0062-Allow-DHCP-Broadcast-packets.patch
```diff
@@ -0,0 +1,74 @@
+From 52c1cff91bab2da916816783f53dc5ee47a9ee13 Mon Sep 17 00:00:00 2001
```
chaitanyagodavarthi, line 1: Why two commits for the same bug fix?

NagConnectus: WIFI5 and WIFI6 kernel code is different since they are running on different linux kernel versions.

NagConnectus:
> Have tested it in the case of vlan?

I don't have vlan setup currently, sure I'll test that case

patches/0062-Allow-DHCP-Broadcast-packets.patch
chaitanyagodavarthi: Yes, but the two patches can be in the same commit. push it on a branch from wifi5, and pull the same branch and add the wifi6 code.

NagConnectus:
> > Have tested it in the case of vlan?
> 
> I don't have vlan setup currently, sure I'll test that case

done

patches/0062-Allow-DHCP-Broadcast-packets.patch
NagConnectus: To detect the DHCP packet there is local structure in the source code. And didn't find any global structure definitions to detect the DHCP packet.

patches/0062-Allow-DHCP-Broadcast-packets.patch
NagConnectus: Yes, that can be doable. But for time being I'm keeping them separate for now.

chaitanyagodavarthi (approved):

ArifAlam (approved):

PR #381, opened 2021-09-17
WIFI-2410: Filter out low signal SSID from SM neighbour report

shimmy568:
Added a check to filter out SSID's with a low signal from out neighbour reports. Removes neighbours that have a RSSI value below the minimum value for that radio type. 

### DIFF
--- a/feeds/wlan-ap/opensync/patches/44-sm_set_min_sig_neighbour_report.patch
+++ b/feeds/wlan-ap/opensync/patches/44-sm_set_min_sig_neighbour_report.patch
@@ -1,0 +1,33 @@
+--- a/src/sm/src/sm_neighbor_report.c
++++ b/src/sm/src/sm_neighbor_report.c
+@@ -43,6 +43,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBI
+ 
+ #define MODULE_ID LOG_MODULE_ID_MAIN
+ 
++#define MIN_RSSI_2G -83
++#define MIN_RSSI_5G -85
++
+ typedef struct
+ {
+     bool                            initialized;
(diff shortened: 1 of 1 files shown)


feeds/wlan-ap/opensync/patches/44-sm_set_min_sig_neighbour_report.patch
```diff
@@ -0,0 +1,32 @@
...
++                "Remove %s %s neighbor due to weak signal {bssid='%s' ssid='%s' rssi=%d min_rssi=%d chan=%d}\n",
++                radio_get_name_from_cfg(radio_cfg_ctx),
++                radio_get_scan_name_from_type(scan_type),
++                scan_entry->bssid,
++                scan_entry->ssid,
++                scan_entry->sig,
++                min_sig,
++                scan_entry->chan);
++            continue;
++        }
```
netexp-ravi, line 28: Seems OK. But, cosmetically I would like to add the new filter to the existing one.

We already have a filter to remove entries with rssi_signal==0. We can add more filters to the same if case like below.

Current filter in the same function,
        /* Filter neighbor with 0 SNR */
        if (scan_entry->sig == 0) {

Add one more filter to the existing if case. Please also update the comment.
 if ((scan_entry->sig == 0) || (scan_entry->sig < min_sig))

shimmy568: Fixed, also added some constants for the min values just so they aren't hardcoded directly in there. Probably overkill but hey...

netexp-ravi (approved):

ArifAlam (approved):

PR #383, opened 2021-09-21
WIFI-4209: Fix Maverick SSID

ArifAlam:
Apply maverick ssid config after timeout.


RickSommerville (approved):
radio_maverick() is called from periodic_task() which no longer uses the reload_config flag to trigger a call to system("reload_config").  As a result, the Maverick SSID is configured (written to UCI) but not loaded/activated.  This fix adds the required reload operation.  Seems correct and low risk because will only occur if the Cloud is NOT connected which in turn means nothing else should be modifying the configuration.

NagConnectus (approved):

PR #384, opened 2021-09-27
wifi-4421: Fix client disconn when in DFS channel

chaitanyagodavarthi:
On detecting Radar on the DFS channel the radio switches
the channel to backup channel. However, the opensync tries
to set it back to the configured channel every 30 seconds, which
which would fail since the channel is in a radar NOP period, this
leads to dropping of clients.
Fix by adding a check for the configured channel's availability
before trying to change to that channel.

### DIFF
--- a/feeds/wifi-ax/hostapd/patches/f00-021-chan-switch-check-dfs-channel-availability.patch
+++ b/feeds/wifi-ax/hostapd/patches/f00-021-chan-switch-check-dfs-channel-availability.patch
@@ -1,0 +1,29 @@
+Index: hostapd-2020-07-02-58b384f4/src/ap/ubus.c
+===================================================================
+--- hostapd-2020-07-02-58b384f4.orig/src/ap/ubus.c
++++ hostapd-2020-07-02-58b384f4/src/ap/ubus.c
+@@ -1016,12 +1016,18 @@ hostapd_switch_chan(struct ubus_context
+ 		css.freq_params.center_freq1,
+ 		css.freq_params.center_freq2);
+ 
+-	if ((chan->flag & HOSTAPD_CHAN_RADAR) &&
+-	    ((chan->flag & HOSTAPD_CHAN_DFS_MASK)
+-	     != HOSTAPD_CHAN_DFS_AVAILABLE)) {
+-		wpa_printf(MSG_INFO, "%s: DFS chan need CAC", __func__);
+-		hostapd_switch_chan_dfs(iface, &css);
+-		return UBUS_STATUS_OK;
++	if (chan->flag & HOSTAPD_CHAN_RADAR) {
++		if ((chan->flag & HOSTAPD_CHAN_DFS_MASK) ==
++		     HOSTAPD_CHAN_DFS_UNAVAILABLE) {
++			wpa_printf(MSG_WARNING, "DFS chan %d not Available",
++				   chan->chan);
++			return UBUS_STATUS_NOT_SUPPORTED;
++		} else if ((chan->flag & HOSTAPD_CHAN_DFS_MASK)
++			!= HOSTAPD_CHAN_DFS_AVAILABLE) {
++			wpa_printf(MSG_DEBUG, "DFS chan do CAC");
++			hostapd_switch_chan_dfs(iface, &css);
++			return UBUS_STATUS_OK;
++		}
+ 	}
+ 
+ 	for (i = 0; i < hapd->iface->num_bss; i++) {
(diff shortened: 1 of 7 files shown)


NagConnectus (approved):
Looks good to me!

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -330,6 +341,8 @@ static bool radio_state_update(struct uci_section *s, struct schema_Wifi_Radio_C
 			SCHEMA_SET_INT(rstate.channel, chan);
 		else
 			SCHEMA_SET_INT(rstate.channel, blobmsg_get_u32(tb[WDEV_ATTR_CHANNEL]));
+			radio_fixup_set_primary_chan(rstate.if_name,
```
RickSommerville, line 344: Indentation is off or missing {} on else cause.

Why update this on the radio_state_update() instead of on a config operation?

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
```diff
@@ -452,8 +465,15 @@ bool target_radio_config_set2(const struct schema_Wifi_Radio_Config *rconf,
...
 	if (changed->channel && rconf->channel) {
-		blobmsg_add_u32(&b, "channel", rconf->channel);
-		rrm_radio_rebalance_channel(rconf);
+		/* If radio channel config has changed then rebalance channel,
+		 * if not, that means the channel state changed due to a 
+		 * radar detection and shall remain in the backup channel */
+		if (is_cloud_channel_change(rconf) == 0) {
+			if(rrm_radio_rebalance_channel(rconf) == 0) {
+				blobmsg_add_u32(&b, "channel", rconf->channel);
+				radio_fixup_set_primary_chan(rconf->if_name, rconf->channel);
```
RickSommerville, line 474: Why does radio_fixup_set_primary_chan() call depend on whether the the rrm_radio_rebalance_channel() worked?

I guess it gives us retries if the rebalance fails.  However, I think the call to radio_fixup_set_primary_chan() in the state update probably prevents this.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
chaitanyagodavarthi: We update it in the config too, but in case there is no config change yet like when bootup, the current config is got from the uci using tb[WDEV_ATTR_CHANNEL]. So our fixup variable and uci wireless config are same always. The new channel coming from the vif config  is compared with this before setting it.

feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/radio.c
chaitanyagodavarthi: We compare the fixup chan with the radio config channel , if they are different we do a rebalance_channel. So in case the rebalance_channel fails, the opensync would retry to set the channel and since the fixup chan and radio config chan are different even now , it proceeds to rebalance_channel. The radio state update only copies the channel value from wireless uci.

feeds/wifi-ax/hostapd/patches/f00-021-chan-switch-check-dfs-channel-availability.patch
```diff
@@ -0,0 +1,31 @@
...
+-		return UBUS_STATUS_OK;
++	if (chan->flag & HOSTAPD_CHAN_RADAR) {
++		if ((chan->flag & HOSTAPD_CHAN_DFS_MASK) ==
++		     HOSTAPD_CHAN_DFS_UNAVAILABLE) {
++			wpa_printf(MSG_WARNING, "DFS chan %d not Available",
++				   chan->chan);
++			return UBUS_STATUS_NOT_SUPPORTED;
++		}
++		
++		if ((chan->flag & HOSTAPD_CHAN_DFS_MASK)
```
RickSommerville, line 23: Could line 23 "if" be an "else" to line 16 "if"?  More readable, since second if just negates the first,

RickSommerville (approved):

feeds/wifi-ax/hostapd/patches/f00-021-chan-switch-check-dfs-channel-availability.patch
chaitanyagodavarthi: done

ArifAlam (approved):

PR #387, opened 2021-09-29
Create ngx.yml

mrinalmadhukar-inventum:


blogic:
this patch has formal errors. please read https://openwrt.org/submitting-patches on how to format patches.

ngx.yml
```diff
@@ -0,0 +1,13 @@
+---
+profile:  qcom_ap-dk01.1-c1
+target:  ipq40xx
+subtarget: generic
+description: Build image for the ngx
+include:
+  - ucentral-ap-light
+  - ucentral-ap
```
blogic: you cannot have both

ngx.yml
```diff
@@ -0,0 +1,13 @@
+---
+profile:  qcom_ap-dk01.1-c1
```
blogic: this is dk01, hence the file has to be called qcom_ap-dk01.1-cs1.yml. I am not sure why we have a build profile for a reference kit.

PR #390, opened 2021-10-04
Wifi-4469. Set registered flag before config QDSS

netexp-ravi:
The QDSS config takes long time to complete and puts the kernel
thread to freeze for a while. This lets the kernel run other
threads which leads to country code command execution without
the ATH11K_FLAG_REGISTERED flag set, which inturn leads to
failure of Country code configuration.

Updating the ATH11K_FLAG_REGISTERED flag before calling QDSS
config seems to help driver configure the Country code successfully.

### DIFF
--- a/feeds/wifi-ax/mac80211/patches/qca/226-ath11k-set-registered-flag-before-qdss-config.patch
+++ b/feeds/wifi-ax/mac80211/patches/qca/226-ath11k-set-registered-flag-before-qdss-config.patch
@@ -1,0 +1,11 @@
+diff -Naur a/drivers/net/wireless/ath/ath11k/core.c b/drivers/net/wireless/ath/ath11k/core.c
+--- a/drivers/net/wireless/ath/ath11k/core.c	2021-09-30 09:44:22.861587609 -0400
++++ b/drivers/net/wireless/ath/ath11k/core.c	2021-09-30 09:45:18.361000898 -0400
+@@ -847,6 +847,7 @@
+ #endif
+ 	ath11k_hif_irq_enable(ab);
+ 
++	set_bit(ATH11K_FLAG_REGISTERED, &ab->dev_flags);
+ 	ath11k_config_qdss(ab);
+ 
+ 	if (ab->hw_params.ce_fwlog_enable) {


feeds/wifi-ax/mac80211/patches/qca/226-ath11k-set-registered-flag-before-qdss-config.patch
```diff
@@ -0,0 +1,11 @@
+diff -Naur a/drivers/net/wireless/ath/ath11k/core.c b/drivers/net/wireless/ath/ath11k/core.c
+--- a/drivers/net/wireless/ath/ath11k/core.c	2021-09-30 09:44:22.861587609 -0400
++++ b/drivers/net/wireless/ath/ath11k/core.c	2021-09-30 09:45:18.361000898 -0400
+@@ -847,6 +847,7 @@
+ #endif
+ 	ath11k_hif_irq_enable(ab);
+ 
++	set_bit(ATH11K_FLAG_REGISTERED, &ab->dev_flags);
```
chaitanyagodavarthi, line 8: the ath11k_core_qmi_firmware_ready is also called form ath11k_core_reconfigure_on_crash, and in that function the ATH11K_FLAG_REGISTERED is not set. So this flag needs to be set only when the ATH11K_QMI_EVENT_FW_READY event comes and not in the other case. 

Also, is it possible to set the number of attempts it tries to get the qdss fw, so we can limit the number of tries to 1 and it wouldn't cause much delay.

feeds/wifi-ax/mac80211/patches/qca/226-ath11k-set-registered-flag-before-qdss-config.patch
```diff
@@ -0,0 +1,11 @@
...
+ 	ath11k_config_qdss(ab);
```
chaitanyagodavarthi, line 9: Also, please add a detailed comment in the code on why this change is required once my above post is addressed.

feeds/wifi-ax/mac80211/patches/qca/226-ath11k-set-registered-flag-before-qdss-config.patch
netexp-ravi: ATH11K_FLAG_REGISTERED is necessary for the Country code setting. Setting it in ath11k_core_qmi_firmware_ready() from ath11k_core_reconfigure_on_crash() does not seemed to cause any problem, and in-fact seems necessary.

netexp-ravi: Regarding the retries comment. There does not seemed to be multiple retries. Once it tries from Kernel which if fails, it takes User's help to load the file. In both the modes, it is tried only once.

PR #405, opened 2021-11-04
qosify: speedup builds by using host clang/llvm

aparcar:
* Add `host_dependencies` config option to `gen_config.py` script to handle potential host dependencies.
* If available use clang/llvm from host instead of building it which speeds up image generation by some 30 minutes
* Add qosify profile
* Switch to use qosify profile instead of installing package directly


aparcar:
@blogic this technically works but the [bpf backport](https://github.com/Telecominfraproject/wlan-ap/blob/main/backports/0029-bpf-headers-add-a-package-with-kernel-headers-for-eb.patch) is incomplete. It doesn't match up with what we have in the OpenWrt [master branch](https://github.com/openwrt/openwrt/commits/master/include/bpf.mk).

The latest patch from Felix solves this inside the build system if manually downloaded a *pre-packaged* `llvm-bpf` build. This could be the cleanest solution.

aparcar:
Merged in the next branch.

PR #406, opened 2021-11-05
Allow explicit warnings for missing host dependencies. 

aparcar:
The generic warning list could also be used for other warnings later on.


blogic:
merged, Thanks !

PR #411, opened 2021-11-17
WIFI-3512: Fixed stats formula when client reconnects

shimmy568:
Made it so that if the client reconnects (which resets the stats) it doesn't use the difference between the new and only value for the rate, it just uses the new value on its own.

Just went with the naive solution since I couldn't figure out how to access the data for "NL80211_STA_INFO_CONNECTED_TIME" in "nl_call_param" given the params that are passed into "target_stats_clients_convert". Knowing this info we could probably get a more accurate taking into account the shorter collection period for that client since it wasn't connected the entire period.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
@@ -157,13 +157,26 @@ bool target_stats_clients_convert(radio_entry_t *radio_cfg, target_client_record
 	client_record->stats.rssi       = data_new->stats.rssi;
 	client_record->stats.rate_tx    = data_new->stats.rate_tx;
 	client_record->stats.rate_rx    = data_new->stats.rate_rx;
-	client_record->stats.bytes_tx   = data_new->stats.bytes_tx   - data_old->stats.bytes_tx;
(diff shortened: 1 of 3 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
```diff
@@ -148,6 +148,14 @@ bool target_stats_clients_get(radio_entry_t *radio_cfg, radio_essid_t *essid,
         return ret;
 }
 
+uint64_t update_stats_feild(uint64_t new_value, uint64_t old_value) {
```
RickSommerville, line 151: field not feild please.

Needs to be all or none.  If one count has been reset, they've all been reset.  Therefore test all and then either subtract all or not.

In stat_nl80211.c/nl80211_assoclist_recv(), add NL80211_STA_INFO_CONNECTED_TIME to the stats_policy.  Then add it to the stored data target_client_record_t and use it to make the decision.

It doesn't have to be added to the app_client_record_t (would be a mess and require a Cloud change as well).

RickSommerville (commented):
Explained how to do "not naive" in the comments

RickSommerville (approved):

PR #414, opened 2021-11-27
wifi-ax: batman-adv: Add fixes from openwrt-routing 21.02 branch

ecsv:
The batman-adv for non-802.11ax devices is a fork from the openwrt-routing 21.02 branch. The private copy is used because the QCA kernel is extremely old and doesn't match the kernel from OpenWrt 21.02. And the mac80211 version (and its backports/compat code) is not fully compatible with the official one.

The batman-adv updates from the official openwrt-routing 21.02 branch must be regularly, manually cherry picked.


ecsv:
Superseded by #432

PR #415, opened 2021-12-01
Wifi-5995. Fix for CSA failure and invalid chan/BW config

netexp-ravi:
A channel change request from RRM while a reload_config is in
progress cause the new CSA attempt fail. AP updated the UCI
no matter if the CSA attempt failed or passed. This leads to
the AP not re-trying the CSA again.

Furthermore another issue is seen when an invalid channel/BW
combination is configured. Solution is to update the Bandwidth
as per channel change request is made via CSA.

Example: BW already configured for 40MHz, and then the channel
change attempt is made to 165. The channel change code takes care
of downgrading the BW to 20MHz while sending the request down to
the driver, but did not update the same in the UCI config.
UCI config reads as 165/40. A reloadconfig thereafter causes this
invalid combination 165/40 attempt which leads to radio stay disabled.


yashvardhan-netex (commented):
Let's discuss more on this. The patch would negatively impact the hitless provisioning we want to achieve when rrm does a channel rebalance.

yashvardhan-netex (approved):
Looks good!

PR #416, opened 2021-12-05
Add minim support for linksys_ea6350-v4 and add openwifi support for Gl-iNet B1300 and Motorola MH7020 

alrooney:
1. Add Minim feed and unum package
2. Add minim/unum package (minim agent) to Linksys EA635-v4
3. Add OpenWiFi support for Gl-iNet B1300 and Motorola MH7020


alrooney:
We will resubmit without the minim components and just device support.

PR #420, opened 2021-12-14
Wifi-6165. Handling durationMs from Chan_info events

netexp-ravi:
Chan_info events use the same container as BSS_info events,
hence the durationMs from Chan_info events need special handling.
Adding a separate counter to track the old and new durationMs
from Chan_info events and adding the differential duration to the
old Bss_info duration.

### DIFF
--- a/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
+++ b/feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
@@ -278,36 +278,61 @@ bool target_stats_survey_get(radio_entry_t *radio_cfg, uint32_t *chan_list,
 
 bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t scan_type,
 				 target_survey_record_t *data_new, target_survey_record_t *data_old,
-				 dpp_survey_record_t *survey_record)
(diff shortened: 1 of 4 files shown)


feeds/wlan-ap/opensync/src/platform/openwrt/src/lib/target/src/stats.c
```diff
@@ -287,27 +288,48 @@ bool target_stats_survey_convert(radio_entry_t *radio_cfg, radio_scan_type_t sca
...
-		data_new->chan_busy += data_old->chan_busy;
+		LOGD("target_stats_survey_convert: for_report:%d, scan_type:%d, chan:%d,"
+			"oldSurveyP:%p, newChanDur:%llu, oldDur:%llu, newBusy:%llu, oldBusy:%llu,"
+			" new_nf:%u, old_nf:%u, scInt:%d\n",
+			for_report, scan_type, data_new->info.chan, data_old, data_new->duration_ms,
+			data_old->duration_ms, data_new->chan_busy, data_old->chan_busy,
+			data_new->chan_noise, data_old->chan_noise, scan_interval);
+
+		if ((data_new->duration_ms <= 150) &&
+			(data_new->duration_ms <= (uint64_t)scan_interval)) {
```
yashvardhan-netex, line 299: As discussed, the above condition would not work as expected. We have to come up with a different solution.

PR #422, opened 2021-12-16
WIFI-6180: Fix cmdm crash

ArifAlam:
Fix cmdm crash when LED definition is not present in board.json.


chaitanyagodavarthi (approved):
Looks good.

PR #429, opened 2022-02-14
Update APK to uvol/uxc compatible version

aparcar:
This commit replaces the incomplete apk build and instead adds the version 3.0.0_pre0 which handles uxc/uvol containers.


aparcar:
Opened against wrong branch.

PR #430, opened 2022-02-14
README: fix linksys_ea8300 example

aparcar:
The profile is no longer called just `ea8300`.

Signed-off-by: Paul Spooren <mail@aparcar.org>


blogic:
merged, Thanks !

PR #432, opened 2022-02-26
batman-adv: Synchronize with OpenWrt 21.02.2 package

ecsv:
OpenWrt 21.02.2 was just released and contains a couple of changes to the batman-adv package. The (QCA) non-11ax wlan-ap devices will receive this automatically when the routing feed is updated. But for (QCA) 11ax devices, this has to be imported manually because the wlan-ap setup script is not patching the routing repository to get support for the QCA Linux 4.4 kernel fork and instead uses a private copy of batman-adv.


blogic:
merged, Thanks !

PR #435, opened 2022-03-09
Inventum Ap ngx firmware version udaya-1.1

PankajNatrajan:
Inventum Ap profile and changes is added after review by olivier ,please use the readme or staging-udaya-1.0 for compiling instructions.


blogic:
@PankajNatrajan please be a little more careful with PRs

PR #439, opened 2022-04-07
Fixed typo in cig_wf610d profile name

shimmy568:
I noticed there was a typo in the profile name for the cig_wf610d AP profile name. So this PR is just to rename the file and change all references to cig_wf160d


blogic:
merged, Thanks !

PR #442, opened 2022-04-13
usteer: fix uchannel.uc script

stintel:
Once the uchannel.uc script sets its status to waiting, it becomes
defunct. The next invocation of the script will compare uptime to
state.uptime, and if the difference is less than 12h, it will change
state.uptime to the current uptime and return. As the script runs every
5 minutes, the difference will always be less than 12h.

Fix this by comparing uptime against state.changed rather than
state.uptime.

While at it, rename state.uptime to state.executed, to be more
descriptive.


blogic:
merged, Thanks !

PR #445, opened 2022-05-04
ucentral-schema: Fix ucentral_apply.sh

bmastbergen:
Pass the config to ucentral.uc

Fixes: 0b9087bd440c ("ucentral: development update")
Signed-off-by: Brett Mastbergen <brettm@minim.com>


blogic:
this is an old script ... please use following call signature " /usr/share/ucentral/ucentral.uc /etc/ucentral/examples/default.json" 

I'll remove the old libexec script tomorrow, sorry forgot it even existed

PR #447, opened 2022-05-19
ipq807x: fix dts for gl-ax1800

gl-zhaojianhui:
Signed-off-by: Jianhui Zhao <jianhui.zhao@gl-inet.com>


blogic:
merged, Thanks !

PR #448, opened 2022-05-23
ipq807x: add wireguard support

gl-zhaojianhui:
The wireguard source in feeds/ipq807x/ipq807x/files/drivers/net/wireguard is from

https://git.zx2c4.com/wireguard-linux-compat/snapshot/wireguard-linux-compat-1.0.20211208.tar.xz

The changes for wireguard source only delete one file: wireguard/Makefile

Except add the wireguard source, there is a bug in qca 5.4 kernel, also fixed.

Signed-off-by: Jianhui Zhao <jianhui.zhao@gl-inet.com>


blogic:
thanks ! we will be adding wireguard suport to the uCentral data model this week aswell, will test this patch once that has happened

blogic:
I have merged a different version of the patches.

PR #466, opened 2022-07-27
ipq807x: Fix installation of FTBFS gcc 5.5 patch

ecsv:
The toolchain path will not automatically be copied to OpenWrt. So the build will still fail on recent Debian unstable version. Instead, it must be applied during the setup.py process via the normal patch procedure.

Reported-by: Marek Lindner <marek.lindner@plasma-cloud.com>
Fixes: 6f6b40a94ff6 ("toolchain/gcc: fix building gcc 5.5 on newer systems")


blogic:
merged, Thanks !

PR #468, opened 2022-08-12
sk_buff's cb should be cleared in the entry point for both incoming and outgoing  packets at each layer.

venkatchimata:
At bridge, it is already handled in the outgoing path,
but not in the incoming path. We have seen cases where proxyarp_replied was 1
on sk_buffs coming from the WLAN driver and they were getting dropped in forwarding path.


blogic:
merged, Thanks !

PR #470, opened 2022-09-01
Add Multi Band Operation support to ApNOS

stintel:
These are mostly backports from OpenWrt where the feature has been tested.
We need to bump usteer before backporting the hostapd changes, ad the hostapd changes drop the wnm_disassoc_imminent ubus method, which might still be used in the current version of usteer.


stintel:
Closing in favour of #471

PR #473, opened 2022-09-07
ipq807x: Fix build of kmod-br-netfilter

ecsv:
QCA is still backporting the old OpenWrt hack for BR_HOOK to newer kernels. This change was identified as a source for bugs and also caused performance problems in certain scenarios.

And since OpenWrt is now building the bridge as part of the kernel and the br-netfilter code as module, the build of the QCA kernel fork will also fail because the kernel binary is depending on a module:

    net/built-in.o: in function `BR_HOOK':
    net/bridge/br_private.h:929: undefined reference to `br_netfilter_run_hooks'
    net/bridge/br_private.h:929: undefined reference to `br_netfilter_run_hooks'
    net/bridge/br_private.h:929: undefined reference to `br_netfilter_run_hooks'
    net/bridge/br_private.h:929: undefined reference to `br_netfilter_run_hooks'
    net/bridge/br_private.h:929: undefined reference to `br_netfilter_run_hooks'
    net/built-in.o:net/bridge/br_private.h:929: more undefined references to `br_netfilter_run_hooks' follow

Just getting rid of this patch (as OpenWrt did) is therefore the best option.


blogic:
merged, thanks

PR #476, opened 2022-09-12
WIFI-10739 OpenWiFi periodically (typically 120 seconds) issues nl80211

venkatchimata:
call to get associated client list from the AP. The associated client list was protected by a mutex lock. Somehow it was causing tx/rx issues sometimes, speciically there was jitter in the video calls. Changing the lock to rcu_read lock from mutex. This resolved the issue and now there is no jitter in the calls.


venkatchimata:
Opening a clean pr

PR #477, opened 2022-09-13
OpenWiFi periodically (typically 120 seconds) issues nl80211

venkatchimata:
call to get associated client list from the AP. The associated client list was protected by a mutex lock. Somehow it was causing tx/rx issues sometimes, speciically there was jitter in the video calls. Changing the lock to rcu_read lock from mutex. This resolved the jitter issue.

Tested thoroughly with the following procedure.

1. Connect 3 clients
2. A script running dumping clients in an infinite loop
3. Continuously disconnect / connect one client to see if there is any crash. No crash was observed.


greearb:
Testing locking races is very difficult to do with great confidence, do you have reason to believe that an rcu read lock is actually sufficient?  I think at the least the patch should go to the linux-wireless mailing list for review by Johannes and others.

venkatchimata:
I see in other places say tx.c where the sta list is protected by rcu read lock. Thought we could use it here too.

PR #491, opened 2022-11-04
[cosmetic] base-files: fix foreign sysupgrade detection

pepe2k:
Backups are gzipped, include the missing `z` parameter in `tar` call. While at it, make `grep` call quiet (add `q` parameter).
This is more a cosmetic change, relevant only to locally/manually made sysupgrade.

Should cleanly apply to `main` and `next`.


blogic:
merged, thanks !

PR #500, opened 2022-11-21
base-files: make EAP104 switch ports configurable

ArifAlam:
Fixes WIFI-11709


blogic:
merged, Thanks !

PR #507, opened 2022-12-14
Radius NAS IP improvements WIFI-11935

venkatchimata:
1. Add NAS IP Address support to hostapd configuration
2. Check for stations in all interfaces across radios while handling client disconnect request from Radius


blogic:
@venkatchimata thanks, kindly send the schema patch via email

PR #516, opened 2023-01-20
ucentral-client: health: remove redundant check for zero value

Cahb:
config_get is issued with a default (120) parameter, which makes a check for 0 a redundant.
It can be safely removed, as the check for "[ "$interval" -eq 0 ]" is always non-zero.

Fixes 396e2bd06cbe ("ucentral-client: cleanup health parameter")

Tested on virtual Wlan-AP img (OpenWRT): no syntax error occured / executed OK.

Signed-off-by: Oleksandr Mazur <cahbua@gmail.com>


blogic:
merged to my local branch, will be in next shortly, Thanks !

PR #517, opened 2023-01-21
Staging wifi 12182 support oap100e

tanyasingh0902:
WIFI-12182: Support Edgecore OAP100e


patches/ipq40xx/0019-ipq40xx-support-oap100e.patch
```diff
@@ -0,0 +1,477 @@
...
++	$(call Device/FitImage)
++	$(call Device/UbiFit)
++	DEVICE_VENDOR := Edgecore
++	DEVICE_MODEL := OAP100e
++	SOC := qcom-ipq4019
++	BLOCKSIZE := 128k
++	PAGESIZE := 2048
++	IMAGES := nand-sysupgrade.bin
++	DEVICE_DTS_CONFIG := config@ap.dk07.1-c1
++	DEVICE_PACKAGES := ipq-wifi-edgecore_oap100e kmod-usb-acm kmod-usb-net kmod-usb-net-cdc-qmi uqmi
```
blogic, line 452: @tanyasingh0902 i think the BDF being selected here is not part of the PR

tanyasingh0902: @blogic OAP100 and OAP100e have the same BDF and other configs except for the GPIO for external antennas for 5G radio OAP100e. Please suggest what we can do here.

blogic: @tanyasingh0902 make the devbice select the correct BDF in that case. the current patch selects a non-existing BDF I think.

tanyasingh0902: @blogic the BDF of OAP100 is the same as OAP100, so will it be acceptable if we change 
`+	DEVICE_PACKAGES := ipq-wifi-edgecore_oap100e kmod-usb-acm kmod-usb-net kmod-usb-net-cdc-qmi uqmi`
to select OAP100 bdf
`+	DEVICE_PACKAGES := ipq-wifi-edgecore_oap100 kmod-usb-acm kmod-usb-net kmod-usb-net-cdc-qmi uqmi`
?

blogic: @tanyasingh0902 yes, that would be ok

PR #518, opened 2023-01-24
uboot-envtools: Fix syntax on number of sectors

netexp-ravi:
ProblemStatement: On EAP104, "fw_setenv" fails with the error "MTD erase error on /dev/mtd10: Invalid argument"
RootCause: uboot-envtools expects hex format for the number of sectors while the config file (/etc/fw_env.config) contains a decimal number format for the number of sectors. This leads to out-of-boundary size calculation within the function "environment_end()".
Solution: Update the scripts to write a hex value instead of a decimal value in the config file.

Reference for the expected format from the config file: Please check the function "get_config" in the file "fw_env.c", where the function uses sscanf to read the parameters from the config file.

Fixes: WIFI-12189
Signed-off-by: Ravi Vaishnav <ravi.vaishnav@netexperience.com>


blogic:
@netexp-ravi kindly explain in the description why this is needed. as in why the code fails to do the dec vs hex conversion

netexp-ravi:
> 

I updated the commit log which explains my findings. Please let me know if more information is needed.

blogic:
merged, Thanks!

PR #526, opened 2023-02-24
ucentral-event: wifi-12324 5g band is showing as unknown

Satya-sc:
ucentral-event: wifi-12324 5g band is showing as unknown


blogic:
merged, Thanks !

PR #527, opened 2023-02-28
[for main/next]: bluetooth backports for LE-only HCI controllers

pepe2k:
This is for **main** (+ **release/v2.9.0**) and **next** branches.

The most important fix is a backport from 4.9 for LE-only HCI Bluetooth controllers: https://github.com/torvalds/linux/commit/39385cb5f3274735b03ed1f8e7ff517b02a0beed. The three additional backports from second commit were found missing during some initial testing.

Due to the fact that the kernel 4.4 used currently in OpenWiFi lacks many more Bluetooth subsystem related fixes, we should consider backporting the whole stack from more up to date version, with e.g. the backports project.


blogic:
merged, Thanks !

PR #539, opened 2023-03-31
ipq807x: fix qca-nss-clients feed

ArifAlam:
Update qca-nss-clients feed source url to fix build failure.


jaspreetsachdev:
@blogic does this look good to you?  I would like to get the builds going now.

PR #542, opened 2023-04-24
staging-wifi-12512-support-eap104-ethernet-phy-check

tanyasingh-ec:
for https://telecominfraproject.atlassian.net/browse/WIFI-12512


blogic:
@tanyasingh-ec please enhance the commit message to explain that the tool does and why it is needed

PR #544, opened 2023-05-08
uspot/spotfilter: improvements and bugfixes

f00b4r0:
This PR:
- goes on :)


feeds/ucentral/uspot/files/usr/bin/captive
```diff
@@ -17,6 +17,68 @@ function interfaces() {
...
+	}
+
+	if (!parse_bool(uci.get('uspot', name, 'configure_spotfilter')))
+		exit(0);
+
+	let uspot = uci.get_all('uspot', name);
+	if (!uspot)
+		fail('Cannot load uspot config for "' + name);
+
+	let device_macaddr = uci.get('network', uspot.interface, 'device');
```
f00b4r0, line 41: Technically, this makes it possible to setup spotfilter redirect to an interface that isn't part of the same network as the wlan devices. Is this ok?
Shouldn't spotfilter be hooked to a network interface (as in, config/network interface) instead of specific devices?
It would then make it trivial to have uspot running on a router and have downstream dumb APs bridged to that interface.

PR #549, opened 2023-06-01
Staging qsdk ath12 wifi 12617 support oap100e

tanyasingh-ec:
For WIFI-12617: To support OAP101 E without 6GHz interface


blogic:
merged, Thanks !

PR #555, opened 2023-07-13
WIFI-12743 CIG WF-186w : Fixed the vlan issue when add two vlan inter…

Kenshi0000:
…face into one bridge


blogic:
merged, Thanks !

PR #559, opened 2023-08-05
Let qca thermald support IPQ50xx platform and has cooling function to…

wingate5678:
… limit the duty cycle of tx queues


wingate5678:
This is a enhance function for thermal management for now is use for IPQ50xx serial only , it configure the duty cycle of thermal device to suspend all data tx queues and limit the data transmission to control the temperature

blogic:
merged, Thanks !

PR #564, opened 2023-08-11
Add the corresponding software content for the <wallys-dr5018> hardwa…

wallystech-Gg:
…re device


wallystech-Gg:
Add the corresponding software content for the <wallys-dr5018> hardware device

PR #565, opened 2023-08-18
Modify Wifi Board Data for Cybertan EWW631 Serial

wingate5678:
For Cybertan EWW631 serial optimize the wifi board data


wingate5678:
update new one

PR #567, opened 2023-09-01
add Xunison D50 and Q50 devices

outbackdingo:


blogic:
can you move these to v5.4 kernel directly instead of v4.4 please

PR #568, opened 2023-09-04
WIFI-12907: cig-wf186w fix mpsk dynamic vlan

phwhite:


phwhite:
This is still raw and under test, let's not merge yet

feeds/ipq807x/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -65,6 +65,7 @@ qcom_setup_interfaces()
 		;;
        cig,wf186w)
 		ucidef_add_switch "switch0" "4:wan" "1:lan" "2:lan" "3:lan" "0:lan" "6@eth0"
+		ucidef_set_interface_wan "eth0"
```
Kenshi0000, line 68: Any side effect since eth0.XXXX was used for WAN interface with switch vlan enabled ?

phwhite: @Kenshi0000 I did this so that the existing ucentral-event `vlad_add` functionality works.   To be honest, I'm not sure if it's broken other functionality yet or not.   Are you aware of how this is used in the system?

phwhite: @Kenshi0000 It turns out this part of the patch wasn't needed, since I am overwriting the event WAN interface in the `99-ucentral-event` file.   

However, just to note -- removing this results in /etc/board.json ending up with still broken/wrong config.    @blogic @Kenshi0000 is this OK, or should something be done to properly fix /etc/board.json and thus the downstream users of this info (like the original ucentral-event)?
```
        "network": {
                "wan": {
                        "device": "eth0.1",
                        "protocol": "dhcp",
                        "macaddr": "cc:cf:83:ea:be:80"
                },
                "lan": {
                        "device": "eth0.2",
                        "protocol": "static"
                }
        },
```

blogic: @phwhite no that is fine, the board.json is used to derive the default OpenWrt config

phwhite:
@blogic Fixed the VID logic, and tested to make sure everything is still working.    Also rebased it on top of current next.  This should be good to go now.

blogic:
merged, Thanks !

PR #574, opened 2023-09-18
WIFI-12942 CIG WF186W: Fixed the certificate lost issue with dual ima…

Kenshi0000:
…ge enabled


blogic:
the 186w fix is already merged, please rebase on latest next branche

PR #575, opened 2023-09-18
WIFI-12941 CIG wf660ag: support both EMMC and SPI NOR+EMCC H/W

Kenshi0000:
Using one firmware to support the board which has EMMC only and SPI NOR+EMMC equipped

### DIFF
--- a/feeds/ipq807x/ipq807x/base-files/lib/upgrade/platform.sh
+++ b/feeds/ipq807x/ipq807x/base-files/lib/upgrade/platform.sh
@@ -84,11 +84,17 @@ emmc_do_upgrade_bootconfig() {
 	fi
 
 	for part in "0:BOOTCONFIG" "0:BOOTCONFIG1"; do
-		local emmcblock=$(find_mmc_part $part)
-		echo erase ${emmcblock}
-		dd if=/dev/zero of=${emmcblock} 2> /dev/null
-		echo update $emmcblock
-		dd if=/tmp/bootconfig of=${emmcblock} 2> /dev/null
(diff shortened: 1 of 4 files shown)


feeds/ipq807x/ipq807x/base-files/lib/upgrade/platform.sh
```diff
@@ -84,11 +84,17 @@ emmc_do_upgrade_bootconfig() {
 	fi
 
 	for part in "0:BOOTCONFIG" "0:BOOTCONFIG1"; do
-		local emmcblock=$(find_mmc_part $part)
-		echo erase ${emmcblock}
-		dd if=/dev/zero of=${emmcblock} 2> /dev/null
-		echo update $emmcblock
-		dd if=/tmp/bootconfig of=${emmcblock} 2> /dev/null
+               local emmcblock=$(echo $(find_mtd_chardev $part) | sed 's/^.\{5\}//')
```
blogic, line 87: emmcblock -> mtdchar ?

feeds/ipq807x/ipq807x/base-files/lib/upgrade/platform.sh
```diff
@@ -84,11 +84,17 @@ emmc_do_upgrade_bootconfig() {
 	fi
```
blogic, line 84: emmc_do_upgrade_bootconfig is now incorrect call it something wf660 specific please

feeds/ipq807x/ipq807x/base-files/lib/upgrade/platform.sh
Kenshi0000: > emmcblock -> mtdchar ?

just use same name for mtd cha and emmcblock.

PR #589, opened 2023-10-05
[next/trivial] Revert "feeds: ucentral: include BlueZ 5.66"

pepe2k:
This reverts commit 8111566f35c33de51f41eece000ef4daf4916313.

After switching to 'OpenWrt 23.05' release tree we no longer need separate BlueZ 5.66 package as it's available in upstream feed.


blogic:
merged thanks !

PR #590, opened 2023-10-05
[next] ipq807x: ipq5018: fix EdgeCore EAP104 pinctrl configuration

pepe2k:
Drop `phy_led_pins` node which duplicates configuration for `gpio46`, used in different place (`led_cloud` node). This solves below error:

```
pin GPIO_46 already requested by 1000000.pinctrl; cannot claim for leds
```


blogic:
merged thanks !

PR #593, opened 2023-10-25
update EWW631-A1 and EWW631-B1

wingate5678:
update EWW631-A1 BDF and EWW631-B1 (to 512M Memory)


wingate5678:
missing some files

PR #594, opened 2023-10-25
EWW631 A1 update and B1 support 512M memory

wingate5678:
update both wifi BDF for A1 and B1 and EWW631-B1 support 512M memory


wingate5678:
change another account to push code

PR #595, opened 2023-10-26
Update EWW631-B1 to support 512M Memory

jackcybertan:
(1) EWW631-B1 to support 512M Memory
(2) update A1 and B1 BDF for FCC rule


blogic:
merged, Thanks !

PR #597, opened 2023-11-14
[WIFI-13155] add OptimCloud D series support

bacnh85:


blogic:
merged, Thanks !
please send another PR purging all remains of the xunison board support

PR #600, opened 2023-11-24
ipq807x: add support for OptimCloud D series

bacnh85:
Since QC only supports IPQ5018 + QCN9000 combination perfectly in kernel 4.4, thus remove OptimCloud support in ipq807x_v5.4 and migrate to ipq807x.


blogic:
merged, Thanks !

PR #603, opened 2023-11-29
5.4 Kernel Upgrade for HFCL ipq60xx targets

esha-aggarwal18:
This PR covers 5.4 upgrade for HFCL targets hfcl_ion4xe, hfcl_ion4xi, hfcl_ion4x, hfcl_ion4x_2


kish-shukla:
@esha-aggarwal18 changing the ion4xi profile image path as name incorrectly mentioned!

kish-shukla:
hfcl_ion4x has an incorrect package requirement, it should use ipq60xx not ipq807x in the profile

PR #605, opened 2023-12-04
[for next] feeds: mcu: umcumgr: update to latest HEAD

pepe2k:
This is for **next** branch.

This updates the `umcumgr` CLI tool, bringing various set of fixes and most importantly, support for `Texas Instruments serial bootloader` (enabled by default). Additionally, support for `confirm` and `test` commands was disabled by default as we don't make use of it (all supported boards use either `direct-XIP` wit multiple slots or `single-slot` configuration).

Cheers,
Piotr


blogic:
merged, Thanks !

PR #607, opened 2023-12-05
removed legacy xu dtsi files

outbackdingo:
removes XU series dtsi files from tree


bacnh85 (approved):
LGTM!

blogic:
please fold the 4 commits into a single one

PR #612, opened 2023-12-07
uspot: fix uam CHAP password and challenge

ArifAlam:
Add missing last byte in uam CHAP password and challenge

Fixes WIFI-13178


blogic:
merged, Thanks !

PR #617, opened 2024-01-19
cybertan_eww631-b1: Fix LAN port mappings

phwhite:
The LAN port mappings were reversed, causing the port with LAN1 label to be reported as LAN3 in uCentral


blogic:
merged, Thanks !

PR #619, opened 2024-01-23
dvlan: dhcpsnoop fix

phwhite:
When a DVLAN is initialized, it creates a new eth0.XXX interface. This interface didn't have DHCP snooping enabled on it, causing uCentral to not have the connecting clients IP address info


blogic:
merged, Thanks !

PR #621, opened 2024-01-26
hapd, ucentral-event: fix 802.1x with dynamic VLAN

phwhite:
This adds a workaround to fix an issue with 802.1x + DVLANs on platforms
where LAN ports are through an integrated switch (swconfig).

Netifd is tracking the wired ports as part of a bridge-vlan: either a
static one, or 4090 for the default untagged bridge.  When hostapd
authorizes the wired port, netifd is automatically adding this bridge
vlan as PVID untagged to the port.  The vlan_add event then adds the
dynamic VLAN as untagged to the same port.  The result is that the
port is operating on the PVID bridge vlan, and not the dynamic VLAN.
Fixing this in netifd is going to be complex and take time, so this
change includes a workaround.   When a wired client is authorized
using a dynamic VLAN, ucentral-event takes the following actions:
   - Remove the bridge VLAN from the port
       bridge vlan del dev <port> vid <bridge-vlan>
   - Modify the dynamic VLAN to PVID
       bridge vlan add dev <port> vid <dynamic-vlan> pvid untagged

### DIFF
--- a/feeds/ucentral/ucentral-event/files/ucentral-event
+++ b/feeds/ucentral/ucentral-event/files/ucentral-event
@@ -34,6 +34,7 @@ let boot_file = "/tmp/booted";
 let ucentral_running = false;
 let pending_events = [];
 let relay = {};
+let net_config = {};
 
 
 function config_load() {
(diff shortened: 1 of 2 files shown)


blogic:
merged, Thanks !

PR #622, opened 2024-01-26
eap-104: fix dualboot: reset bootcount

phwhite:
The bootcount wasn't getting reset to 0 upon a successful boot,
resulting in falling back to the previous firmware version after three
reboots of the AP.


blogic:
merged, Thanks !
also fixed mount_certs ...

PR #630, opened 2024-02-29
WIFI-13253: Update edgecore eap111 dts file to fix eth0 and eth1 MAC …

tanyasingh-ec:
…addr assignment


blogic:
merged, Thanks !

PR #634, opened 2024-03-09
[Next] Cybertan EWW631-A1 and EWW631-B1 both support v5.4

wingate5678:
Cybertan EWW631-A1 and EWW631-B1 both support v5.4


blogic:
merged, Thanks !

PR #642, opened 2024-03-25
netifd: add psk2-radius encryption support

bacnh85:


bacnh85:
As `wallys-dr6018-v4` is not in `next` so I cherry-pick from `main` and combine with a patch to `netifd` to support `psk2-radius` encryption.

bacnh85:
This issue is fixed by other commit: https://github.com/Telecominfraproject/wlan-ap/commit/1ec29f67054aa1be5b6e5e6f3bb7b58149135927

PR #644, opened 2024-03-29
hostapd: WISPr bandwidth for psk2-radius

ArifAlam:
Get bandwidth limits from WISPr RADIUS AVPs for RADIUS ACL.

Fixes WIFI-13550


blogic:
merged, Thanks !

PR #654, opened 2024-04-12
[WIFI-13576] certificates: add write_certs_sqsh helper script for JFFS2 fstype

bacnh85:
Mount yuncore AX820 as `squashfs`


feeds/tip/certificates/files/usr/bin/mount_certs
```diff
@@ -10,15 +10,25 @@ check_certificates
 mkdir -p /certificates /etc/ucentral/
 mtd=$(find_mtd_index certificates)
 
-if [ "$(head -c 4 /dev/mtd$mtd)" == "hsqs" ]; then
-	mount -t squashfs /dev/mtdblock$mtd /certificates
-else
+rootfstype=$(cat /proc/cmdline | sed 's/.*rootfstype=\([^ ]*\).*/\1/')
+case "$rootfstype" in
+squashfs)
```
blogic, line 15: this will break 90% of boards

patches/0078-ramips-yuncore-ax820-increase-certificate-parition-to-512kB.patch
```diff
@@ -0,0 +1,47 @@
...
+
+diff --git a/target/linux/ramips/dts/mt7621_yuncore_ax820.dts b/target/linux/ramips/dts/mt7621_yuncore_ax820.dts
+index b2f55b9be0..570a34d65e 100644
+--- a/target/linux/ramips/dts/mt7621_yuncore_ax820.dts
++++ b/target/linux/ramips/dts/mt7621_yuncore_ax820.dts
+@@ -120,12 +120,12 @@
+ 			partition@90000 {
+ 				compatible = "denx,uimage";
+ 				label = "firmware";
+-				reg = <0x90000 0xf60000>;
```
blogic, line 20: will break all ax820 boards currently deployed

bacnh85: All ax820 boards currently deployed - `certificates` folder is not mounted properly - instead at `overlay` partition.

`check_certificates` is called at first, so if there is certs inside `/certificates` folder then no mount mechanism run, thus this will not break ax820 currently deployed.

feeds/tip/certificates/files/usr/bin/mount_certs
bacnh85: Let me update the PR with updated code to cover this case.

blogic: you will have to verify this on every supported board

patches/0078-ramips-yuncore-ax820-increase-certificate-parition-to-512kB.patch
blogic: changing the size will break, people currently deploy squashfs 64k block in the partition

bacnh85: got it - let me revise back to 9k and test more on this device.

PR #664, opened 2024-05-27
WIFI-13428 : Fix traffic disruption for MPSK clients for yuncore_fap655,cigwf186w,eap104

joydeepbenison:
Fixes WIFI-13428 : traffic disruption for FAP655,CIG_WF186w,EAP-104 for MPSK Clients when VLAN id of the Lan port is modified . This Commit configures the dynamic VLANs on the switch of fap655,eap104,cig_wf186w again after reset.

### DIFF
--- a/patches/0072-WIFI-13428-Fix-Traffic-Disruption-for-MPSK-clients-f.patch
+++ b/patches/0072-WIFI-13428-Fix-Traffic-Disruption-for-MPSK-clients-f.patch
@@ -1,0 +1,111 @@
+From da1d005199f6485c37aed3ae15faf1b9ab79d88d Mon Sep 17 00:00:00 2001
+From: joydeepbenison <joydeep.ghosh@benisontech.com>
+Date: Mon, 27 May 2024 13:34:25 +0530
+Subject: [PATCH] WIFI-13428 Fix:Traffic Disruption for MPSK clients for
+ yuncore_fap655
+
+Fix Traffic Disruption for MPSK clients for yuncore_fap655 when
+vlan-id of the lan port is modified.Any change in swconfig resets the
+internal switch of FAP655 due to which all the dynamically added vlans
+for MPSK clients are lost. Configure the Dynamic Vlans for MPSK Clients
+again after switch reset.
+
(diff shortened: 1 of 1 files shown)


patches/0072-WIFI-13428-Fix-Traffic-Disruption-for-MPSK-clients-f.patch
```diff
@@ -0,0 +1,111 @@
...
++
++debug() {
++	logger -t HOTPLUG-DVLAN "$*"
++}
++
++if [ "${INTERFACE:0:2}" != "up" ]; then
++	exit 0
++fi
++
++case "$(board_name)" in
```
phwhite, line 40: I'd really prefer if we can keep per-model config like this in a central place, rather than having various parts of the system that need updating upon adding new models.    I believe all this info is available in `/etc/board.json` and, to some degree, in `/etc/config/network/` and `/etc/config/event` -- I'd suggest trying to read it from `board.json` as the single source of truth though.   What do you think?

Here's snippet from `board.json` of WF-186W

```
	"switch": {
		"switch0": {
			"enable": true,
			"reset": true,
			"ports": [
				{
					"num": 4,
					"role": "wan"
				},
				{
					"num": 0,
					"role": "lan"
				},
				{
					"num": 1,
					"role": "lan"
				},
				{
					"num": 2,
					"role": "lan"
				},
				{
					"num": 3,
					"role": "lan"
				},
				{
					"num": 6,
					"device": "eth0",
					"need_tag": false,
					"want_untag": false
				}
			],
```

phwhite (commented):
Initial review looks OK, however I've added a comment checking if we can use `/etc/board.json` for the switch info, rather than defining per-model info in the newly added script?

patches/0072-WIFI-13428-Fix-Traffic-Disruption-for-MPSK-clients-f.patch
blogic: @phwhite patch imho uses the right place in the chain of events to fix the issue, yet the code is not optimal. I plan to clean this up tomorrow.

PR #672, opened 2024-06-20
WIFI-13858: Update board data file for EAP102 to support DFS channels…

tanyasingh-ec:
… correctly on India (IN) country code


blogic:
merged, Thanks !

PR #676, opened 2024-07-01
ieee8021x / hostapd: Fix 802.1x + DVLAN issues

venkatchimata:
ieee8021x
---------
1. Handle link_up events and update hostapd config
2. For certains scenarios, we need to remove and add instead of reload (reload did not work). Consider the following scenario - Say on CIG 186w as an example eth0.4086 interface exists with some non-ieee8021x config. Push ieee8021x config. In general the flow is that reload_config is called followed by invocation of services (from ucentral-schema) Services inovation does n't wait until the configi reloaded ie in this context ieee8021x service is invoked much before the network interfaces are recreated. That is not correct. To handle this, we capture link-up events and remove the existing interface (in hostapd as shown below) and add again
3. For swconfig platforms, the names contain a dot. Handle that gracefully in ubus_unsub_object while adding hostapd interface
4. Add better logging using log.syslog

hostapd
---------
In case of swconfig switches, the basename of the interface should be based on the last dot. Earlier it was done based on the first dot, which would result in incorrect basename. For example if the interface name is eth0.4087 then the vlan->ifname would be eth0.4087.  (A dot at the end) . Before this patch, the basename was returned as eth0. It should be eth0.4087

Also fixed the return code by adding a default value of 0 and removed an unncessary check for if_add before ubus add call.


blogic:
merged, Thanks !

PR #678, opened 2024-07-04
wireguard: update package hash

ArifAlam:


malcolmmadsheep:
@ArifAlam hi, thanks for the commit! Do you have any idea, why the hash was changed? Because I see that the file is 2 years old, so at first glance it shouldn't be changed and I'm wondering what is the reason.

blogic:
merged, Thanks !

blogic:
@malcolmmadsheep I compared the content of the 2 downloads and it is identical

PR #679, opened 2024-07-08
mac80211: Update tx stats correctly in case of AP mode

venkatchimata:
In the backports driver the tx stats are updated in ieee80211_8023_xmit. However in AP mode the packets are transmitted in ieee80211_8023_xmit_ap. ieee80211_8023_xmit is not hit in case of AP mode. Update the stats in ieee80211_8023_xmit_ap path too.


blogic:
merged, Thanks !

PR #680, opened 2024-07-10
ipq50xx: Adjust ubootenv_size for ipq50xx

venkatchimata:
ubootenv_size is computed dynamically similar to ipq807x. This resolves the issue where fw_setenv fails with the following error.

MTD erase error on /dev/mtd13: Invalid argument
Error: can't write fw_env to flash


blogic:
merged, Thanks !

PR #681, opened 2024-07-10
eap104: Update the environment variable bootcount properly for EAP104

venkatchimata:
In the bootcount script there was no entry for Edgecore EAP104 to set the bootcount. This commit adds support for Edgecore EAP104 in bootcount script to set the environment variable correctly


blogic:
merged, Thanks !

PR #682, opened 2024-07-15
ieee8021x: Fix connectivity issues with a non-vlan based event

venkatchimata:
Non-vlan based ieee8021x client was not able get IP Addess through the AP. The reason was that ieee8021x.uc subscriptions could n't subscribe hostapd events as the subscribe functions are called before creating the interfaces.

To handle this, invoke hapd_subscriber.subscribe after the interface is created. Also dont delete the interface from ifaces as ifaces should always contain the configured ports (from uci config) per ieee8021x.uc design. Deleting interfaces from ifaces will lead to ieee8021x.uc crashes.


blogic:
merged, Thanks !

PR #683, opened 2024-07-17
Update wpa2-radius.json

firasshaari:
Fix configurations for the WPA2 Enterprise example with 802.11r enabled for fast transitioning.


blogic:
merged, Thanks !

PR #687, opened 2024-07-26
firewall4: Enable the HW offloading for Firewall4

ramesh-benison:
Fixes the throughput issue while traffic run from LAN to WAN ports on AP yuncore ax820 model using wired clients.

Test scenario:
WiredClient[192.168.1.67] <======> (LAN)APX820(WAN) <=====> WiredClient[192.168.100.103]


ramesh-benison:
Throughput reached ~938 Mbits/sec(UL or DL) between LAN/WAN ethernet ports using iperf3.

blogic:
this should be rendered out via the data-model

ramesh-benison:
> this should be rendered out via the data-model

I am not able to find any data-model  in wlan-ucentral-schema which actually uses or override cases for firewall defaults section.
So the firewall defaults section is always intact.

blogic:
provided alternative fix to benison

PR #691, opened 2024-07-31
add sonicfi rap630w 211g

CybertanYingWei:
add sonicfi rap630w 211g

### DIFF
--- a/feeds/mediatek-sdk/mediatek/mt7981/base-files/lib/upgrade/platform.sh
+++ b/feeds/mediatek-sdk/mediatek/mt7981/base-files/lib/upgrade/platform.sh
@@ -32,6 +32,9 @@ platform_do_upgrade() {
 		fi
 		nand_do_upgrade "$1"
 		;;
+	sonicfi,rap630w-211g)
+		/tmp/nand_sonicfi_rap630w_211g.sh "$1"
+		;;		
 	esac
 }
 
@@ -46,6 +49,13 @@ platform_check_image() {
 		nand_do_platform_check "$board" "$1"
 		return $?
 		;;
+	sonicfi,rap630w-211g)
+		[ "$magic" != "73797375" ] && {
+			echo "Invalid image type."
+			return 1
+		}
+		return 0
+		;;		
 	esac
 
 	return 0
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [ 'cig_wf186h', 'cig_wf186w', 'cig_wf188n', 'cig_wf196', 'cig_wf189', 'cybertan_eww631-a1', 'cybertan_eww631-b1', 'edgecore_eap101', 'edgecore_eap102', 'edgecore_eap104', 'edgecore_eap105', 'edgecore_eap111', 'edgecore_oap101', 'edgecore_oap101-6e', 'edgecore_oap101e', 'edgecore_oap101e-6e', 'hfcl_ion4xe', 'hfcl_ion4xi', 'hfcl_ion4x', 'hfcl_ion4x_2', 'hfcl_ion4x_3', 'hfcl_ion4xi_w', 'hfcl_ion4x_w', 'indio_um-305ax', 'sercomm_ap72tip', 'udaya_a6-id2', 'wallys_dr5018', 'wallys_dr6018', 'wallys_dr6018-v4', 'yuncore_ax820', 'yuncore_ax840', 'yuncore_fap640', 'yuncore_fap650', 'yuncore_fap655' ]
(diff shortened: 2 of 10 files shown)


feeds/mediatek-sdk/mediatek/mt7981/base-files/lib/upgrade/platform.sh
```diff
@@ -32,6 +32,9 @@ platform_do_upgrade() {
 		fi
 		nand_do_upgrade "$1"
 		;;
+	sonicfi,rap630w-211g)
+		/tmp/nand_sonicfi_rap630w_211g.sh "$1"
```
blogic, line 36: what is this needed for ?

CybertanYingWei: The main reason is that when upgrading the image, we cannot use the original nand.sh because we are currently using a private structured image. Therefore, we need to write a separate private upgrade script.

CybertanYingWei: The main reason is that when upgrading the image, we cannot use the original nand.sh because we are currently using a private structured image. Therefore, we need to write a separate private upgrade script.

blogic: can we use a default image inside TIP images ?

CybertanYingWei: No, we need to use a third-party boot code during the production process to support switching between dual images.

blogic: please provide details of why this is needed

CybertanYingWei: Rap630w_211G is using a third-party boot code, so its image structure needs to be compatible with the boot code. Therefore, it needs to make modifications to the sysupgrade process in order to work properly. Therefore, it cannot use the nand.sh function that comes with tip. At the same time, it needs to take care not to affect other tip model projects. Therefore, it has separated the private nand.sh and used the private script during the sysupgrade process.

CybertanYingWei: @blogic   Hello, may I ask if our push request has been approved this time? Thank you!

blogic: /tmp/nand_sonicfi_rap630w_211g.sh is not part of the patch and the resulting image would not be upgradable

blogic: also there is no valid Signed-off-by line in the commit

CybertanYingWei: @blogic :
In the file (patches/0072-add-sonicfi-rap630w-211g.patch),the patch will create the file (package/base-files/files/lib/upgrade/sonicfi/nand_sonicfi_rap630w_211g.sh) and add some codes for the 'package/base-files/files/sbin/sysupgrade'.

+board=$(board_name)
+case "$board" in
+	sonicfi,rap630w-211g)
+	cp /lib/upgrade/sonicfi/nand_sonicfi_rap630w_211g.sh /tmp/nand_sonicfi_rap630w_211g.sh
+	chmod 777 /tmp/nand_sonicfi_rap630w_211g.sh
+	;;
+esac
+

When the 'sonicfi rap630w-211g' Router is upgrading the image.Sysupgrade script will copy the 'nand_sonicfi_rap630w_211g.sh' from the '/lib/upgrade/sonicfi' directory to the '/tmp/' directory.

blogic: @CybertanYingWei why does it need to be copied to tmp ? the file should also be located in feeds/mediatek/mediatek/base-files/....

CybertanYingWei: Because when performing the step of writing NAND flash, the system has already umounted all partitions, and the path (/lib/upgrade/) is no longer accessible. Therefore, it is necessary to copy it to '/tmp/' in advance in order to access the files inside normally.

blogic: @CybertanYingWei openwrt has a facility for this. you can set
RAMFS_COPY_BIN=''
RAMFS_COPY_DATA=''
inside platform.sh

CybertanYingWei: Okay, I will modify the code again, thank you!

CybertanYingWei: I have revised the relevant code as per your request. Now, please close this push request and push the new code again. Thank you!

PR #702, opened 2024-08-14
WIFI-13879: Support GPS on OAP101 series

tanyasingh-ec:


feeds/ipq807x_v5.4/hostapd/files/gps_lock
```diff
@@ -0,0 +1,62 @@
...
+
+  [ -f "$GPS_DUMP_TMP" ] && rm -f "$GPS_DUMP_TMP"
+  [ -f "$GPS_DUMP" ] && rm -f "$GPS_DUMP"
+
+  # change tty port to GPS
+  echo 0 > /sys/class/gpio/ble_enable/value
+  # change stty for GPS output
+  stty raw -echo -echoe -echok -F /dev/ttyMSM1 115200
+
+  i2cset -f -y 0 0x74 0x07 0x12
```
blogic, line 41: what chip is located at 0x74 ? and what do these i2c calls do ?

tanyasingh-ec: @blogic The 0x74 is the CPU's gpio expander. i2c controls the GPIO expander pin to reset the GPS. 

i2c commands should only be used in the gps_reset script and not gps_lock script. Should I modify the gps_lock script in this branch itself or create a fresh branch with correct gps_lock script?

blogic: @tanyasingh-ec I have just pushed and alternate version of the patch to the next branch. with the patch applied I am able to call "ubus call gps info"

tanyasingh-ec: @blogic As I understand, /etc/init.d/ucentral-gps shall enable/disable GPS using the ble_enable and gps_reset gpio. How about the GPS NMEA data? Should I create a new patch based on your patch https://github.com/Telecominfraproject/wlan-ap/commit/386612b34cb16153695a8bf99d1f477924fa10c7 to get the GPS NMEA data or will it be handled differently?

blogic: my patch configures the UCI for the GPS daemon and it will be started on boot. the only missing piece is to read the gps location via ubus and then pass it on to AFC daemon

tanyasingh-ec: @blogic OK, we will test this and provide a patch to support it then

PR #707, opened 2024-08-27
hostapd: add WPA-PSK key mgmt for psk2-radius

mkwacz:
Using WPA-PSK-SHA256 caused issue with WiFi clients connection which support only WPA-PSK.
Added WPA-PSK to avoid of regression and solving
isssue with clients connectivity.

Fixes: 0a21b9d2 ("hostapd: enable FT-PSK for psk2-radius")


firasshaari:
I tested "wpa_key_mgmt=WPA-PSK WPA-PSK-SHA256". It failed with my iPhone15 Pro and my OnePlus12 phone. The phones will always choose 256

blogic:
merged, Thanks!

PR #721, opened 2024-09-12
KAN-3 RAP630 series support PWM-LED

jackcybertan:
change LED control from GPIO to PWM

base on QCA "ipq5018_wi-fi_access_point_soc_data_sheet.pdf" and "IPQ50XX_SOC_Software_User_Guide.pdf"
add kernel config CONFIG_LEDS_PWM and patch pinctrl-ipq5018.c


jackcybertan:
close this case and re-create by JIRA Issue

PR #725, opened 2024-09-18
Removing 802.11w and SHA256 encryption from PSK-Radius

firasshaari:
Removing 802.11w and SHA256 encryption feature from PSK-Radius for MediaTek based APs


blogic:
merged, Thanks !

PR #736, opened 2024-10-09
WIFI-14202 wifi thermal mitigation daemon

jackcybertan:
In Old APNOS (v3.0.2) with kernel 4.4 has qca-thermal package and is for QCA wifi chip thermal mitigation , but this package was removed in new APNOS, we try to write a similar daemon to monitor Wifi chip temperature and mitigation throughput when temperature is too high


blogic:
merged, Thanks !

PR #738, opened 2024-10-16
Fix mac address assignment for eth0 (wan) and eth1 (lan) interfaces in Edgecore EAP105

tanyasingh-ec:


blogic:
merged, Thanks !

PR #740, opened 2024-10-24
WIFI-14171: Update phy driver for EAP105 LAN port

tanyasingh-ec:
The fix has two parts:
1. Update the RTK PHY driver.
2. During initialization, reset PHY with GPIO and setup PHY registers


blogic:
merged, Thanks !

PR #741, opened 2024-10-30
WIFI-14253 WF196- ARQ114C support on IPQ807X platform

Kenshi0000:


jaspreetsachdev:
@Kenshi0000 have you tested this image and ensured that base connectivity to the controller works? Trying to understand what regression QA needs to do here with this change?

PR #743, opened 2024-11-08
WIFI-14270: Add sleep time of 1 second before sending the command to add GST (NMEA) message to /dev/ttyMSM1 port

tanyasingh-ec:


blogic:
merged, Thanks !

tanyasingh-ec:
> merged, Thanks !

@blogic I still don't see this commit in the next branch. Will it be added to the next branch later?

blogic:
@tanyasingh-ec I just pushed the patch

PR #749, opened 2024-11-26
WIFI-14291: Modify example json config for 'wwan' to follow the design in ucentral-schema

tanyasingh-ec:


blogic:
merged, thanks !

PR #759, opened 2024-12-20
ucentral-event: add channel switch handler

mkwacz:
Fix reporting invalid channel in client.join event after channel switch.

Fixes: WIFI-14336


blogic:
merged, Thanks !

PR #766, opened 2025-01-14
WIFI-14348 RAP630W-312G fix interface mdio bus assign error

wingate5678:
fix the mdiobus to mdio-bus and select the mdio0 as mdio bus in dts


blogic:
merged, Thanks !

PR #767, opened 2025-01-16
WIFI-14350 RAP630W-311G fix wifi data memory region

wingate5678:
For RAP630W-311G: fix wifi data memory region size for kernel 5.4

IPQ5018 data      14M > 13M 
QCN6122_1data 16M > 13M
QCN6122_2ata   16M > 13M


blogic:
merged, Thanks !

PR #772, opened 2025-02-04
WIFI-14404: Enable dual boot support on EAP105 and certificate loading on dual boot for EAP104 and OAP101 series

tanyasingh-ec:


blogic:
merged, Thanks !

PR #784, opened 2025-03-14
WIFI-14452 Crash on Cybertan rap7110c-341x with TIP dbdc168 image

wingate5678:
old Wifi BDF was based on 12.2.CSU1 and it will crash when 6G is enanble
change to 12.5.CS BDF with RF's modify


blogic:
merged, Thanks !

PR #787, opened 2025-03-21
WIFI-14488 RAP7110C-341X enable PWM LED config.

wingate5678:
1) Enable PWM-LED in kernel configuration
2) Add Realtek RT82xx PHY with clkout disable and aldps enable


blogic:
merged, Thanks !

PR #796, opened 2025-03-28
ipq807x_v5.4/mac80211: ath11k: fix peer addition/deletion error on st…

mkwacz:
…a band migration

Fix removing rhash when station is roaming between radios and apply logic from opensource driver during adding/removing peer.

Fixes: 7374c39d ("ath11k: fix STA roaming between radios")
Fixes: WIFI-14457


mkwacz:
@blogic today discussed with @ArifAlam. Currently we don't observe FW crashes so seems that we can merge changes.
@ArifAlam please only confirm that you apllied changes from this PR in your setup.

ArifAlam:
Confirmed.  Applied this patch and didn't observe crash so far.

blogic:
merged, Thanks !

PR #798, opened 2025-04-02
Support HaLow on EAP112

SebastianHuang-ec:
1. porting MorseMicro HaLow driver to support HaLow on EAP112
2. Only support FCC regulation because of hardware limitation
3. Add /etc/init.d/halow-gpio-reset to initialize HaLow chip in early stage


blogic:
merged, Thanks !

PR #805, opened 2025-04-07
hostapd: proxy arp: fail gracefully

ArifAlam:
Disable proxy arp and continue if snoop interface setup fail.

Fixes WIFI-14507


blogic:
merged, Thanks !

PR #806, opened 2025-04-08
mediatek/hostap: fix roaming ASSOC REJECT

mkwacz:
During roaming validation, an issue was detected with the ASSOC-REJECT event, which had a status code of 55.

To resolve this, the SNonce from the Auth Request Frame was copied, and the random ANonce is now used
only when expired.

Fixes: WIFI-14326


blogic:
merged, Thanks !

PR #807, opened 2025-04-08
WIFI-14018: Allow option 82 DHCP fields to be transparently injected

Kumiko18:
WIFI-14018: Allow option 82 DHCP fields to be transparently injected into client DHCP requests

      1. Added userspace application udhcpinject to inject DHCP option 82 transparently
      2. Added README.md for example usage

Signed-off-by: alex18_huang <alex18_huang@accton.com>

	new file:   feeds/ucentral/udhcpinject/Makefile
	new file:   feeds/ucentral/udhcpinject/README.md
	new file:   feeds/ucentral/udhcpinject/files/etc/config/dhcpinject
	new file:   feeds/ucentral/udhcpinject/files/etc/init.d/dhcpinject
	new file:   feeds/ucentral/udhcpinject/src/Makefile
	new file:   feeds/ucentral/udhcpinject/src/udhcpinject.c
	new file:   feeds/ucentral/udhcpinject/src/udhcpinject.h


phwhite (commented):
@Kumiko18 Hi, one question: I see this is using tc ingress, and that upon setup it checks if the ingress already exists, but for cleanup it seems to bindly cleanup all ingress rules.     Have you had a chance to test this with the rate limiting feature yet? for example, configure a particular rate limit on the same ssid, and confirm that both features work properly?

It might be nessesary to set a paritcular prio for the tc rules inserted, and then only clean those up if other rules exist?

Kumiko18:
@phwhite I made following modification
1. Explicitly specify the priority of tc ingress rule installed by udhcpinject to 32 (higher than dhcp-relay)
2. Delete tc ingress rule of priority 32 on reload
3. Don't delete ingress qdisc on reload

and tested two scenario
1. rate-limit enabled on SSID1 and then enable udhcpinject
2. udhcpinject enabled on SSID2 and then enable rate-limit
both features seems to work correctly. I'll commit the change to branch.

phwhite:
Thanks @Kumiko18 your TC changes look good to me

blogic:
merged, Thanks !

PR #808, opened 2025-04-08
WIFI-14519 RAP630W-211G change ethernet label and led behavior

jackcybertan:
change ether port label name & led behavior


blogic:
merged, Thanks !

PR #816, opened 2025-04-14
WIFI-14494 add sonicfi rap650c

jackcybertan:
add SonicFi RAP650C


blogic:
@jackcybertan the patch no longer applies, please rebase it. while at it, please fold all four patches into a single one.

PR #817, opened 2025-04-14
WIFI-14529: "udhcpinject" missing from the package list in ucentral-ap.yml

Kumiko18:
Add package udhcpinject into profiles/ucentral-ap.yml


blogic:
merged, Thanks !

PR #818, opened 2025-04-17
WIFI-14530 RAP750W-311A BDF, LED and calibration modify

wingate5678:
1) fix 5G calibration data address
2) modify 2.4/5G BDF
3) change LED to PWM-LED
4) assign LAN Port MAC


blogic:
merged, Thanks !

PR #819, opened 2025-04-17
qca-wifi-7: hostapd.sh: add missed functionality from ipq807x_v5.4

mkwacz:
The main reason was adding missed functionality for Radius configuration which caused rejecting WiFi clients on authentication level.

Still some changes not included:
* AFC,
* hs20_release,
* multiple_bssid,
* he_co_locate

Fixes: WIFI-14459

### DIFF
--- a/feeds/qca-wifi-7/hostapd/files/hostapd.sh
+++ b/feeds/qca-wifi-7/hostapd/files/hostapd.sh
@@ -126,12 +126,15 @@ hostapd_common_add_device_config() {
 	config_add_int maxassoc
 	config_add_int reg_power_type
 	config_add_boolean stationary_ap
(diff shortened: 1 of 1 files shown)


HenryHaller:
On 4.0.0 (7d31bebdb0caf031053dd3c948a3dc771a46924b) I've noticed these 2 issues that I think this PR may help us with:

- Pre-auth/mac auth wlan always puts client into "default vlan" ie the one configured in the interface. Either something has changed in how we should understand the interface configuration or this is an unexpected behavior.
- Post-auth/mPSK fails, see https://telecominfraproject.atlassian.net/browse/WIFI-14540

blogic:
merged, Thanks !

PR #820, opened 2025-04-18
WIFI-14536: Update Channel Utilization Calculation in RRM

tanyasingh-ec:
- avoid calculated Channel Utilization value to be 'infinity' from cycle_count_delta being '0'


blogic:
merged, Thanks !

PR #824, opened 2025-04-24
WIFI-14560: Update file names from /tmp dir in rrmd init script that need to be removed when rrmd (re)starts

tanyasingh-ec:
There was typo in the file names that had to be removed


blogic:
merged, Thanks !

PR #828, opened 2025-04-28
WIFI-14564: Multiple ssids sometime didn't display dhcp option 82 rules

Kumiko18:
Added check when parsing ssid info retrieved from iwinfo. 
Program will exit if expected interface count and iwinfo entry count mismatch.


blogic:
merged, Thanks !

PR #829, opened 2025-04-29
WIFI-14567-WF189-10G-PHY-code-optimization

Kenshi0000:
1. feeds/qca-wifi-7/ipq53xx/dts/ipq5332-cig-wf189.dts --- add "limit_rtlphy_10g_ablity"  in DTS  , no side effect on other product.
2. feeds/qca-wifi-7/ipq53xx/files-6.1/drivers/net/phy/rtk/rtk_phy.c  --  disable 10G capability if DTS defined limit_rtlphy_10g_ablity , no side effect on other product.
3. feeds/qca-wifi-7/qca-ssdk-qca/patches/0006-qca-ssdk-Fix-10G-rtl-phy-driver-for-c45-mdio-read-wr.patch ---  revert the last 0006-qca-ssdk-Fix-10G-rtl-phy-driver-for-c45-mdio-read-wr.patch and based on 0005 patch, only touch rtl826xb_phy.c for CIG 189 only. -- no side effect on other product.


blogic:
merged, Thanks !

PR #834, opened 2025-05-13
WIFI-14582-WF189-Save-crash-log-into-pstore

Kenshi0000:
Reserver ramoops memory for saving crash log, it will be uploaded to cloud


blogic:
@Kenshi0000 could you instead patch the ipq5332-memory.dtsi, that will enable pstore on all devices and not just the wf-189

Kenshi0000:
John,
Since the start system RAM in different product maybe different, so I only put it on 189 dts.
reg = <0x0 0x49c00000 0x0 0x50000>;

blogic:
merged, Thanks !

PR #836, opened 2025-05-14
WIFI-14509-CIG-WiFi7-WF-672A-bring-up

Kenshi0000:
1. bring up wf672A board (SFP 10G, POE 2.5G, 2 bands or 3 bands RF)
2. add drivers rtl8221d for realtek 8221d 2.5G phy (support POE)
3. add drivers lsm303agr ,tmp103 ilps22qs for 3D , temperature  and pressure sensor
4. add cig-wifi-mode-sw for switching radio to 2 bands(2.4G + 4x4 5G) or 3 bands(2.4G +2x2 5G + 2x2 6G)


Kenshi0000:
John,
Just for talking risk.
1. It's  a module which need to be built-in in build profile, so no impact on other vendor if kmod-cig-wifi-mode-sw is not defined in .yml
2. This module will search specific partition named RF_SWITCH, so it will not touch other partition.
3. We ever use the dd to do read and write, but to use /proc file to read is more general.

blogic:
merged, Thanks !

PR #837, opened 2025-05-15
qca-wifi-7: Add Zyxel NWA130BE model

Pan-Zyxel:


jimmy-cheng1983:
ap72tip build fail seems that they use the feeds/ipq95xx, not feeds/qca-wifi-7 in the current next of branch.
What should we do? rebase the staging-WIFI-14580-Support-Zyxel-NWA130BE to the main of branch?

blogic:
yes, please use main branch

PR #839, opened 2025-05-16
Swiitch LAN and WAN port for EMPLUS WAP588M

cpchangemplu:
Swiitch LAN and WAN port for EMPLUS WAP588M


blogic:
merged, Thanks !

PR #840, opened 2025-05-22
WIFI-14623-WF188n-WF196-lost-the-certificates-files-after-upgrade

Kenshi0000:
add 188n and 196 to mount_cert script.


blogic:
merged, Thanks !

PR #841, opened 2025-05-22
WIFI-14603-Default-TC-fq_codel-queue-on-GRE-cause-low-GRE-RX-throughput

Kenshi0000:
add a hotplug script to replace fq_codel queue with noqueue.


blogic:
merged, Thanks !

PR #847, opened 2025-05-27
WIFI-14644-AP72TIP-support-dual-image-and-fix-reset-button-issue

mikeDing1107:
Support dual image;
Fix reset button on AP72TIP;
add AP72TIP-v4 into build list;


mikeDing1107:
Resolved conflicts

PR #849, opened 2025-05-30
WIFI-14650-CIG-WF-189-CIG-WF-189H-LAN-and-WAN-Ports-Swapped-After-Upgrading-to-TIP-devel-14a0c2d-Image

Kenshi0000:
it seems there’s a script syntax error which cause the 02_network can NOT be executed successfully when you add it for “Support for CyberTAN RAP750E-S AP”, please take a look.


blogic:
merged, Thanks !

PR #853, opened 2025-06-06
Staging wifi 14668 add thermal protection for emplus wap588m model

cpchangemplu:
For Emplus WAP588M model, the WIFI  thermal protection is a must-do feature


blogic:
merged, Thanks !

PR #861, opened 2025-06-23
Add Support for EMPLUS WAP581

cpchangemplu:
SoC: Qualcomm® IPQ5016
RF Chipset: IPQ5018 Internal FEM @2.4GHz 2T2R
RF Chipset: QCN6102 Internal FEM @5ghz 2T2R
RAM: 512MB DDR3 RAM
NOR flash: 8MB
NAND flash: 128MB
Ethernet: 3x GbE Ethernet ports
2x RJ-45 PoE IN (WAN/LAN1)
1x RJ-45 PoE OUT
Reset Button: 1
Power Source:In: 802.3at PoE in or DC IN 12v
Out: Passive PSE out (54V for 11.23W / 52V for 10.8W / 50V for 10.4W / 48V for 9.98W) LED Indicator: (tri-color) x 1 (GPIO Control)


cpchangemplu:
https://telecominfraproject.atlassian.net/browse/WIFI-14723

venkatchimata (commented):
I am fully not sure, but curious to know how the mac addresses for the wireless interfaces are derived. I see that for other platforms, the wireless MAC addresses are derived from the eth/base mac.

feeds/ipq807x_v5.4/ipq50xx/base-files/etc/init.d/HW-watchdog
```diff
@@ -0,0 +1,56 @@
+#!/bin/sh /etc/rc.common
```
blogic, line 1: please use the kernel based gpio-wdt
https://github.com/Telecominfraproject/wlan-ap/blob/main/feeds/ipq807x_v5.4/ipq60xx/files/arch/arm64/boot/dts/qcom/qcom-ipq6018-cig-wf188.dts#L337

cpchangemplu:
> I am fully not sure, but curious to know how the mac addresses for the wireless interfaces are derived. I see that for other platforms, the wireless MAC addresses are derived from the eth/base mac.

@venkatchimata ,  we are OEM/ODM company,  I can write MAC address at Calibration data (ART partition)during manufacturing process , does it make sense of you?

venkatchimata:
> > I am fully not sure, but curious to know how the mac addresses for the wireless interfaces are derived. I see that for other platforms, the wireless MAC addresses are derived from the eth/base mac.
> 
> @venkatchimata , we are OEM/ODM company, I can write MAC address at Calibration data (ART partition)during manufacturing process , does it make sense of you?

@cpchangemplu I think that should be ok. Thank you.

venkatchimata:
> > I think the LEDs enumeration is missing. Is it not needed?
> > @venkatchimata  , Due to we are OEM/OEM company , so we don't need to deside the default LED on or off during initail boot up , the customer hadle it self. And  i have enumerate gpio-led in the /sys/class/leds,  customer can easily to switch on/of  LED.   Is it ok for you?
> > 

ok. Thank you.

PR #865, opened 2025-06-27
WIFI-14724

jackcybertan:
Support user to control PSE on/off (chip: tps23861) via ucentral config file.


feeds/ucentral/poe/files/05_poe
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
blogic, line 1: Hi Jack,
please convert this file to use the uci-defaults feature
here is an example
--> https://github.com/Telecominfraproject/wlan-ap/blob/main/feeds/ucentral/ucentral-schema/files/etc/uci-defaults/99-ucentral-event
uci-default files get executed once after firstboot/factory-reset and then deleted
you can modify /etc/config/poe directly without worrying about /etc/config-shadow/
once all uci-default scripts were executed, the system will automatically generate the /etc/config-shadow/ folder
this should make this script much smaller and simpler

feeds/ucentral/ucentral-schema/Makefile
```diff
@@ -6,9 +6,9 @@ PKG_RELEASE:=1
 PKG_SOURCE_URL=https://github.com/Telecominfraproject/wlan-ucentral-schema.git
 PKG_MIRROR_HASH:=c641622188b9b378550c21b1dcfa105b08a60e534d7d77ae63001d38a3e95cd8
```
blogic, line 7: please drop this patch, I have already updated the main branch to latest ucentral-schema HEAD

blogic:
merged, Thanks !

PR #866, opened 2025-06-27
WIFI-14615: fix HaLow client cannot access network

SebastianHuang-ec:
add wds_bridge in hostapd config for HaLow radio


blogic:
merged, Thanks !

PR #868, opened 2025-07-02
qca-ssdk: adjust the amplitude level of SGMII for Zyxel NWA130BE

jimmy-cheng1983:
To fix EMI and avoid Ethernet RX stuck after reboot for Zyxel NWA130BE.
Ethernet Rx stuck was sometimes happened when we do stress reboot, need to adjust the amplitude level of SGMII for CPU and QCA8385 side. 
And those settings come from the result that Zyxel HW team and Manufacturer co-work to fine tune for NWA130BE.


blogic:
merged, Thanks !

PR #871, opened 2025-07-08
qca-wifi-7: update WF672A configuration

Justin5763:
*Reduce the i2c frequency to enable the encryption chip to be recognized
	*Add USB xr Serial driver and init gps uart param
	*Factory reset when switching wifi mode
	*SFP gpio should be input mode


blogic:
merged, Thanks !

PR #874, opened 2025-07-09
WIFI-14631 CIG WiFi 7 Product Save crash log into pstore

CarosLiang:
Enable the feature to  Save crash log into pstore on CIG WiFi 7 products: CIG WF189H, CIG WF89W, CIG WF672


blogic:
merged, Thanks !

PR #876, opened 2025-07-09
WIFI-14817 6G CBP fail on 20M BW

CarosLiang:
update  ATH12.5CS wlan firmware for  all vendors to fix 6G CBP fail on 20M BW


blogic:
merged, Thanks !

PR #877, opened 2025-07-10
WIFI-14751: EAP112 LTE to reconnect automatically after reboot

SebastianHuang-ec:
1. Copy modemmanager to feeds/ucentral.
2. Add function to find the device sysfs path when protocol is "wwan".
3. Call ifup when modem is ready for connection.
4. Add trigger to restart modemmanager when network uci is updated.


blogic:
merged, Thanks !

PR #878, opened 2025-07-11
WIFI-14834 change Sonicfi RAP7 series cert. type from squashfs to EXT4

wingate5678:
change Sonicfi RAP7110C-341X, RAP750xx series certificate type from squashfs to EXT4


blogic:
merged, Thanks !

PR #884, opened 2025-07-16
WIFI-14849 When device is booting,it causes lan/wan ports bridge toge…

JackTsai-CBT:
When the device is booting, LAN/WAN ports will be in bridge state for a while, it will cause clients on the LAN side getting IP address from WAN side's DHCP server. It's wrong behavior. This PR is used for fixing this issue.
This situation usually occurs when device uses IPQ5018+QCA8337 switch chip.
I change that mdio driver bring up from kernel to module autoload and seprate lan/wan port to different group in dts file.


blogic:
merged, Thanks !

PR #890, opened 2025-07-21
feeds.default.conf: switch to using git:// instead of https://

phwhite:


phwhite:
> does this not collide with patches/0088-feed.conf.default-use-the-github.com-based-trees.patch ?

Oh, now I'm confused.  This patch isn't present in our tree that was sync'd with upstream main since then, I'll have to investigate why.   Not only that, but that patch is doing the opposite: changing from git:// to https://, however, our tree is already using http:// and I can't find any patches in our patches/ dir that touch this.

Okay, I'll figure out the mixup on our side regarding the missing patch, but it looks like  this change isn't needed if you want to continue using https:// for the feeds

PR #893, opened 2025-07-22
ratelimit: fix rate limiting on WiFi-7 devices

mkwacz:
Rate limiting was not applied on WiFi-7 devices because their hostapd interface names use the phy* prefix instead of wlan*. This patch extends the match pattern to include both wlan* and phy*.

Fixes: WIFI-14884


blogic:
merged, Thanks !

PR #894, opened 2025-07-23
WIFI-14890-CIG-Support-dual-BDF

Justin5763:
*Support dual BDF for WF186H/WF186W/WF189/WF189H/WF189W/WF672A


blogic:
merged, Thanks !

PR #896, opened 2025-07-24
WIFI-14896-support HaLow Mesh (wlan-ap)

ian77chen:
enable HaLow MESH function in feeds/morse/utils/wpa_supplicant_s1g/Config.in

ticket: WIFI-14896


blogic:
merged, Thanks !

PR #897, opened 2025-07-25
WIFI-14904 Enhance PKI enrollment on squashfs (SonicFi RAP6* series)

jackcybertan:
Ensure PKI2.0 certificates are properly backed up when /certificates is mounted as squashfs (read-only). Enhance the persistent storage mechanism to retain PKI2.0 certificates across factory resets.

Following the current enrollment behavior, `operational.ca` and `operational.pem` - though restored to `/certificates` after a factory reset - are not directly used during enrollment. The `est_client` will still download both certificates again as part of the enrollment process. This patch does not modify that behavior, but simply ensures the certificates are persistently stored across resets.

Impacted models: SonicFi RAP6* series.

### DIFF
--- a/feeds/tip/certificates/files/usr/bin/mount_certs
+++ b/feeds/tip/certificates/files/usr/bin/mount_certs
@@ -58,6 +58,42 @@ udaya,a5-id2)
 		[ -n "$mtd" ] && tar xf /dev/mtdblock$mtd -C /certificates
 	fi
 	;;
+sonicfi,rap6*)
+	bootconfig=$(bootconfig_lookup)
+
+	if [ -n "$bootconfig" ]; then
+		mtd_dev=$(find_mtd_index $bootconfig)
+		block_size=$(cat /sys/class/mtd/mtd$mtd_dev/size)
+		mkdir -p /certificates
+		dd if=/dev/mtdblock$mtd_dev of=/tmp/certs.tar bs=$block_size count=1
+		if tar tf /tmp/certs.tar > /dev/null 2>&1; then
+			tar xf /tmp/certs.tar -C /certificates
+		fi
+		rm -f /tmp/certs.tar
+	fi
+
+	if [ ! -f /certificates/cert.pem ] || [ ! -f /certificates/key.pem ]; then
+		mtd=$(find_mtd_index certificates)
+
+		if [ "$(head -c 4 /dev/mtd$mtd)" == "hsqs" ]; then
+			mount -t squashfs /dev/mtdblock$mtd /certificates
+		else
+			[ -n "$mtd" -a -f /sys/class/mtd/mtd$mtd/oobsize ] && ubiattach -p /dev/mtd$mtd
+			if [ -n "$(ubinfo -a | grep certificates)" ]; then
+				[ -e /dev/ubi0 ] && mount -t ubifs ubi0:certificates /certificates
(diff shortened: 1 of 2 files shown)


feeds/tip/certificates/files/usr/bin/mount_certs
```diff
@@ -42,6 +42,42 @@ sonicfi,rap7*)
...
+			mount -t squashfs /dev/mtdblock$mtd /certificates
+		else
+			[ -n "$mtd" -a -f /sys/class/mtd/mtd$mtd/oobsize ] && ubiattach -p /dev/mtd$mtd
+			if [ -n "$(ubinfo -a | grep certificates)" ]; then
+				[ -e /dev/ubi0 ] && mount -t ubifs ubi0:certificates /certificates
+				[ -e /dev/ubi1 ] && mount -t ubifs ubi1:certificates /certificates
+			fi
+		fi
+
+		if mount | grep "/certificates" | grep -qE "squashfs|ubifs"; then
```
blogic, line 72: there has to be a cleaner way of doing this, instead of copying folders around.

feeds/tip/certificates/files/usr/bin/mount_certs
```diff
@@ -69,12 +67,15 @@ sonicfi,rap6*)
...
-			cp -a /certificates/* /tmp/certs_backup/
-			umount /certificates
-			mkdir -p /certificates
-			cp -a /tmp/certs_backup/* /certificates
+		overlay_name="certs_overlay"
+
+		if mount | grep "/certificates" | grep -qE "squashfs|ubifs" && \
+			! mount | grep "/certificates" | grep -q "$overlay_name"; then
+			mkdir -p /tmp/certs_upper /tmp/certs_work /tmp/certs_merged
+			mount -t overlay "$overlay_name" \
```
blogic, line 75: why do we need an overlay_fs here ? there are 2 optiosn
1) it is ubifs, then we do not have to do anything
2) the certs are in a tar file and can be unpacked to the /certificates folder
both options do not require the complexity of using overlayfs

jackcybertan: Thanks for the question. Here's why overlayfs is included:

- squashfs is read-only  
  During enrollment, new PKI 2.0 certificates are written to /certificates before being archived by store_certs.  
  If it's mounted as squashfs, these writes silently fail and won’t be included in the backup tar.  
  Overlayfs makes the path writable and avoids this issue.

- UBIFS is writable, but we kept the overlay logic for consistency  
  While UBIFS doesn't require overlayfs, maintaining a unified logic path avoids branching code based on filesystem type.  
  This simplifies the codebase and reduces future maintenance effort.

blogic:
please checkout the changes I made earlier today for the udaya board
https://github.com/Telecominfraproject/wlan-ap/blob/main/feeds/tip/certificates/files/usr/bin/mount_certs#L48

jackcybertan:
Thanks for the update and for the changes on the udaya board.

Just to confirm — for squashfs-based `/certificates`, is this the direction you'd recommend we follow moving forward?

From what we observed, the updated flow:

- mounts `/dev/mtdblock*` (the certificates partition) to `/mnt`
- copies its contents to `/certificates`
- unmounts `/mnt`
- extracts the PKI 2.0 certificates into `/certificates`

If this is the expected behavior, we’ll update our implementation accordingly to stay aligned.

PR #902, opened 2025-07-30
qca-wifi-7/afc: Use curl instead of uclient to send/receive AFC request/response using hostapd

tanyasingh-ec:
Fixes: WIFI-14427


blogic:
merged, Thanks !

PR #903, opened 2025-07-30
afc: Use either ca-cert or access-token to get authorization in curl …

tanyasingh-ec:
…command for the AFC server

Fixes:WIFI-14427


blogic:
merged, Thanks !

PR #909, opened 2025-08-06
WIFI-14869 Added ramoops support for SonicFi IPQ5332 devices

jackcybertan:
Added ramoops support for SonicFi IPQ5332 devices


blogic:
merged, Thanks !

PR #910, opened 2025-08-08
cloud_discovery: Track and persist discovery method

mkwacz:
Adds support for recording the method used to discover the cloud controller (e.g. DHCP, FLASH, OpenLAN).
The selected method  records the current date and time along with the discovery method into "/etc/ucentral/discovery_state.json". The date is stored as `YYYY-MM-DD` and time as `HH:MM:SS` (24-hour format).

Fixed: WIFI-14966


feeds/tip/cloud_discovery/files/usr/bin/cloud_discovery
```diff
@@ -78,6 +83,19 @@ function gateway_load() {
 	return readjsonfile('/etc/ucentral/gateway.json');
 }
 
+function discovery_state_write() {
+	let t = localtime();
+	let updated = {
```
blogic, line 88: please just use time()/epoc here. it will make it easier for consumers to just have epoch rather than having to convert/parse a string

PR #911, opened 2025-08-12
base-files: boot: add sync after uci-defaults

phwhite:
A scenario was seen where UCI config was not flushed to disk before an AP power-cycle after uci-defaults was completed.  Since these scripts are deleted after being ran once, there is no way to recover without a factory reset.

Adding this sync operation proved to help avoid this situation from happening


blogic:
merged, Thanks !

PR #913, opened 2025-08-14
WIFI-14970 fix qca83xx switch race condition

wingate5678:
ar8327
Each switch has its own task. If we have multiple switches on the AP, there may be a loop variable contention problem in qca_ar8327_sw_mib_task.
In qca-ssdk 12.x version. mutex is not enough to prevent this race condition because each switch has its own mutex.
If our EAP is set for a long enough time, there may be multiple switches ++loop at the same time,
causing the loop variable to increase by 2 at a time. This will skip the == (2 * (priv->ports) check,
and the loop variable will then increase infinitely, causing an out-of-bound write + crash.
 
In RAP630W-311G has two switches, so this race condition may occur:
> swconfig list
Found: switch0 - QCA MP
Found: switch1 - QCA AR8337
 
If change the mib work time from 120 seconds to 1 second, then found the race condition happen

the patch is from qca-ssdk 13.0 version, in 13.0 already fix it.


blogic:
this patch is already in the main branch

PR #917, opened 2025-08-19
EAP111: Disable surge protection mode for AN8801SB PHY driver

SebastianHuang-ec:
For AN8801SB PHY driver, disabling surge protection mode to prevent TX voltage going too high.


blogic:
merged, Thanks !

PR #918, opened 2025-08-20
ucentral-state: Respect LED config before enabling

phwhite:
Ensure that LEDs are configured to be ON before attempting to change their state.

Previously, if the LED was configured to be OFF, it would still enter a double-blink state when the cloud connection was lost, and then switch to solid ON upon reconnection—ignoring the configured OFF state.

This update changes that behavior:
    - If LEDs are configured OFF, they will remain OFF even during cloud disconnection (no double-blink).
    - After temporary state changes (e.g., during factory reset), the LED will return to its configured state (either OFF or ON).


blogic:
merged, Thanks !

PR #920, opened 2025-08-25
Update the ethernet driver to support AN8801 PHY on Edgecore EAP111

SebastianHuang-ec:
Porting the Ethernet driver to support Edgecore EAP111 for both old hardware (with EN8801SC) and new hardware (AN8801SB).


blogic:
merged, Thanks !

PR #926, opened 2025-09-03
ath11k: improve peer deletion timeout error handling and fix memory leak

ArifAlam:


phwhite (approved):

blogic:
looks like the SoB and annotation got dropped from Venkata's patch

venkatchimata:
i think we can add the following.
1. Good commit description on why this patch helps. I added things here why 100msecs help after my debugging session with the QA. https://github.com/Telecominfraproject/wlan-ap/commit/fcc965567f5fe4143038dff133b95b8eda156013#diff-97a9b79422409dd0ae1f363a6c6f5537e18a239fc16d4af6c5fa53ad6d0726f5R22-R47
2. We can ignore the TIMEDOUT event.

PR #937, opened 2025-09-15
WIFI-15130-WF672-Update-FCC-BDF

Justin5763:
qca-wifi-7:Update WF672 FCC BDF, only support 2-band mode


blogic:
merged, Thanks !

PR #939, opened 2025-09-18
cloud_discovery: run est_client enroll before discovering process

mkwacz:
Run getting EST certs always before discovery process.

Fixes: WIFI-15123


blogic:
why is this needed ?

PR #946, opened 2025-09-24
mediatek-sdk: Support AN8801SB PHY for Edgecore EAP112

SebastianHuang-ec:
Update Edgecore EAP112 device tree to support AN8801SB PHY.


blogic:
merged, Thanks!

PR #952, opened 2025-09-30
WIFI-15163-LAN switch exposes single eth1 instead of per-port interfa…

jackcybertan:
Using VLAN to separate LAN  to each physical port for RAP750W-311A.


blogic:
please put the description of the patch intot he patch itself instead of putting it into the PR

PR #953, opened 2025-10-01
ath12k-wifi: update NWA50BE BDF files

jimmy-cheng1983:


blogic:
please add a better commit message tot he patch, you will also need to create and reference a Jira

PR #1010, opened 2026-02-12
Store RADIUS Class per STA instead of EAPOL SM;      preserve Class for ACL/MAC auth and accounting

venkatchimata:
Move storage of RADIUS Class attributes from the EAPOL state machine to
struct sta_info and update all users accordingly. Previously, Class was
kept only in eapol_state_machine->radius_class, which caused Class to be
lost for authentication paths where eapol_sm is not created. As a result,
Accounting messages and PMKSA cache operations could miss the
Class attribute.

This change makes sta_info->radius_class the single source of truth for
RADIUS Class attributes and ensures they are preserved and echoed in
Accounting regardless of whether an EAPOL state machine exists.


feeds/qca-wifi-6/hostapd/patches/t00-013-Allow-Session-Timeout-with-PSK-RADIUS-during-4-way-h.patch
```diff
@@ -0,0 +1,38 @@
+From e6ec62aa2d68e9436daeb4470260a101a06c9213 Mon Sep 17 00:00:00 2001
```
blogic, line 1: I assume all of these 3 patches are backports from upstream ? @ArifAlam

blogic:
will need a rebase. please open a new PR once rebased.

PR #1018, opened 2026-03-06
ucentral-pstore: Disable printing processed lines

mkwacz:
When processing huge console log and dmesg files printing lines hanged execution blocking ucentral-state startup process.

It seems stdout is mainly for debug purposes and lines are printed after we store file, so it is safe to disable print as file is already processed and can be examined.

Fixes: WIFI-15356


blogic:
merged into my 5.0 staging branch

PR #1019, opened 2026-03-12
ucentral-client: fix ucentral insecure mode enabled by default when a…

ArifAlam:
…llow-self-signed is unset

Fixes WIFI-15375


blogic:
merged into next branch, Thanks !

PR #1020, opened 2026-03-13
WIFI-15377 Fix dhcp checksum error and packet drop when strict forwarding is enabled

Kumiko18:
Fix dhcp checksum error.
Current version of dhcp-inject did not calculate L3/4 checksum correctly, 
packet forwarded by dhcp-inject might be dropped by kernel.

Fix packet drop issue when strict forwarding is enabled.
When strict forwarding is enabled, DHCP response/offer might not be forwarded to client by AP correctly


blogic:
merged into next, Thanks !

please verify, I added an additional endianess fix onto of the series.

tanyasingh-ec:
> merged into next, Thanks !
> 
> please verify, I added an additional endianess fix onto of the series.

@SebastianHuang-ec  Please review

PR #1021, opened 2026-03-13
ucentral-client: fix hostname validation

ArifAlam:
- Wire hostname validation enable / disable to hostname_validate flag in gateway.json

- Fix hostname validation check when server cert Subject CN has other attributes like Organization in case of Insta server certs: subject=CN=*.example.com, O=ExampleInc. Fix backported to libwebsockets from: https://github.com/warmcat/libwebsockets/commit/5124ffe9d431ca866ef90cb6f5167a837fdc4840

Fixes WIFI-15384


blogic:
merged, Thanks !

PR #1023, opened 2026-03-17
qca-wifi-7: Add Zyxel NWA55BE model

Pan-Zyxel:


jimmy-cheng1983:
cause the huge change in v5.0, should i patch in v4.2.3 ?

jimmy-cheng1983:
Hi @blogic ,
why did you close this PR?
Or how should we do for the next step?

PR #1027, opened 2026-03-25
Fixed WIFI-15306:certificates partition fails to mount on rap750w-311a. Need to use UBIFS format certificates on NAND flash.

JackTsai-CBT:
Fixed WIFI-15306:certificates partition fails to mount on rap750w-311a. Need to use UBIFS format certificates on NAND flash.


blogic:
merged, Thanks !

PR #1028, opened 2026-03-25
ucentral-ap: add missing udevmand package

mkwacz:


blogic:
udevmand is gone, ucentral-state now handles it natively via netlink

mkwacz:
Yes, but without this change udevmand from ucentral is not available.

PR #1029, opened 2026-03-26
dhcpsnoop: capture DHCP Option 12 hostname and expose via ubus leases…

venkatchimata:
… method

DHCP clients send their hostname in Option 12 (DHCPHOSTNAME) inside Discover and Request packets, not in the server's ACK. Add support for capturing this hostname and associating it with the MAC/IP entry in the cache.

The approach:
- Parse Option 12 from all DHCP packets in dhcpsnoop_parse_ipv4()
- On Discover/Request, stash the hostname in a pending AVL tree keyed by MAC (cache_pending_hostname). This is necessary because the client sends the hostname, but the server's ACK does not echo it back.
- On ACK, look up the pending hostname by MAC and store it alongside the IP in the main cache entry.

Add a new ubus method 'leases' that returns each tracked client as a nested table with 'ip' and 'hostname' (when known) fields:

  {
    "96:f7:20:1e:55:70": { "ip": "192.168.50.86", "hostname": "iPhone" }
  }

The existing 'dump' method is unchanged so current consumers are unaffected. The new 'leases' method is intended for consumers that need the full lease information including hostname.

Fixes WIFI-15393

### DIFF
--- a/feeds/ucentral/udhcpsnoop/src/cache.c
+++ b/feeds/ucentral/udhcpsnoop/src/cache.c
@@ -18,7 +18,17 @@ struct mac {
         struct avl_node avl;
 	uint8_t mac[6];
 	uint8_t ip[4];
+	char hostname[64];
 	struct uloop_timeout rebind;
+};
+
+/* Temporary store for hostnames seen in client DHCP Discover/Request packets.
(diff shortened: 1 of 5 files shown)


blogic:
merged, Thanks !

PR #1030, opened 2026-03-26
WIFI-15392 fix RAP630W-311G compilation error in next branch

JackTsai-CBT:
RAP630W-311G compilation will fail in branch next

tps23861-poe-ctrl.c:188:32: error: implicit declaration of function 'atoi' [-Wimplicit-function-declaration]
  188 |                         port = atoi(optarg);
      |           
TIP upgrade OpenWrt to version 25.12, which imposes stricter compilation checks, the atoi function requires an explicit inclusion of the <stdlib.h> header file. This header needs to be included in tps23861-poe-ctrl.c.


blogic:
merged with an improved patch description, Thanks !

PR #1042, opened 2026-04-08
docker: set WGET_OPTIONS env variable with a timeout of 30s to preven…

tanyasingh-ec:
…t hanging/failure during downloads (especially for Asia region)


blogic:
merged, Thanks !

PR #1044, opened 2026-04-09
cloud_discover: retain last active config on start if gateway.json is invalid

tanyasingh-ec:
Fixes: WIFI-15430

When the gateway config is not valid, prefer the last active config (tracked via /etc/ucentral/ucentral.active symlink) over the default fallback config. Only fall back to ucentral.cfg.0000000001 if no active config symlink exists or the target file is missing.


tanyasingh-ec:
created another PR for staging 4.2.0 release branch

PR #1045, opened 2026-04-09
ucentral-schema: Mixed SAE protocols require PMF optional (1) at mini…

tanyasingh-ec:
…mum, not required (2)

Fixes: WIFI-15388


tanyasingh-ec:
created another PR for staging 4.2.0

PR #1050, opened 2026-04-15
WIFI-15438: Add sonicfi_rap630w_211g support on openWRT25.12(Kernel6.12)

JackTsai-CBT:
Add sonicfi_rap630w_211g configuration and functionality files on openWRT25.12 and Linux Kernel 6.12.


blogic:
1. include/image-commands.mk — globally breaks Build/libdeflate-gzip

  The ifeq (CONFIG_TARGET_..._sonicfi_rap630w_211g,y) wraps both Build/fit-its and Build/libdeflate-gzip inside the else branch. Result: when Sonicfi is the selected device, Build/libdeflate-gzip is undefined
  globally, breaking every other device whose image recipe uses it in the same build.

  Also, Build/fit is defined only for Sonicfi in the ifeq arm — so when Sonicfi is built, only the Sonicfi variant of Build/fit exists. Any other device built alongside it loses the standard Build/fit.

  Fix: do not override at the global scope. Either (a) define Build/fit-sonicfi and use it from the device's IMAGE/sysupgrade.itb, or (b) upstream the -r rootfs hash feature into the standard scripts/mkits.sh and
  drop the override entirely.

  2. package/base-files/files/bin/config_generate — no-op condition

  if [ "board_name"="sonicfi*" ]; then
      LOG_SIZE="64"
  fi

  [ X ] with a single non-empty argument is always true. This always sets LOG_SIZE=64 on every board, overriding the upstream default of 128. Also board_name is the literal string, not the function call, and =
  doesn't do glob matching anyway.

  Fix: case "$(board_name)" in sonicfi,*) LOG_SIZE=64 ;; esac.

  3. 10-wifi-detect — removes ubus call network.wireless retry for every board

   [ "${ACTION}" = "add" ] && {
  +    mt76_generate_macs
       /sbin/wifi config
  -    ubus call network.wireless retry
   }

  That retry call is what kicks netifd to bring up wifi interfaces whose PHYs showed up after the initial wifi config. Dropping it is a regression for every device, not just Sonicfi.

  Fix: keep the ubus call network.wireless retry line. The MAC-programming step is fine to add.

  4. get_boot_param() duplicated verbatim

  Defined in both target/linux/mediatek/filogic/base-files/lib/rap211g-envtools.sh and target/linux/mediatek/filogic/base-files/lib/upgrade/platform.sh. Identical logic, two copies, will drift.

  Fix: define once in rap211g-envtools.sh (or better, a shared lib/functions/boot_param.sh), source from platform.sh.

  5. rap211g-envtools.sh reimplements existing helpers

  nand_find_ubi() and nand_find_volume() already exist in /lib/upgrade/nand.sh (which is sourced on upgrade). The reimplementations duplicate the logic and use slightly different conventions (e.g. ubi_mknod uses
  basename $dir with no quoting).

  Fix: source /lib/upgrade/nand.sh and drop the duplicates.

  6. fw_printenv | grep ethaddr is fragile

  Used in both 10-wifi-detect and 02_network. Would also match an env var named ethaddr1, ethaddrs, myethaddr, etc., producing wrong values silently.

  Fix: fw_printenv -n ethaddr (prints value only, exits non-zero if missing).

  7. 02_network — misuse of ucidef_set_interfaces_lan_wan arguments

  ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" eth1

  The second argument is a WAN interface name. Other entries in the file use wan (a netdev alias) or concrete switch port names. eth1 is unusual — verify it's actually the intended netdev after the DSA switch setup
   on this board. Probably correct, but worth confirming against the DTS (gmac1 phy-handle &phy0, GMII to MT7531 vs. direct).

  8. 02_network writes /etc/fw_env.config from within mediatek_setup_macs()

  echo "/dev/$env_ubivol 0 0x80000 0x80000 1" > /etc/fw_env.config

  Writing fw_env.config is the job of uboot-envtools init, not board.d/02_network. 02_network runs under /etc/uci-defaults/ on first boot and is the wrong place for persistent fw_env.config. Also, mediatek_filogic
  in uboot-envtools already handles this case for the same board — the logic is duplicated.

  Fix: keep it only in uboot-envtools/files/mediatek_filogic; remove from 02_network.

  9. platform.sh — whitespace/syntax issue in fw_setenv.txt generation

  {
      echo "dual_boot.current_slot" "${upgrade_image_slot}"
      echo "dual_boot.slot_${upgrade_image_slot}_invalid" "0"
  } > /tmp/fw_setenv.txt

  echo a b writes a b (space-separated), but fw_setenv -s expects <name><whitespace><value> with tab as the canonical separator — space works in most fw_setenv implementations, but check the one being used here.
  More importantly, the variable names dual_boot.current_slot contain a dot, which u-boot env accepts but may need careful quoting if expanded elsewhere.

  Minor, but verify on real hardware.

  10. platform_check_image() magic check is squashfs-only

  [ "$magic" != "73797375" ] && { echo "Invalid image type."; return 1; }

  73797375 is "sysu" — the start of a sysupgrade.tar. Fine if the device only ships .tar images, but the filogic.mk also defines IMAGE/sysupgrade.itb with external-static-with-rootfs. If that image is ever flashed
  directly, the magic check rejects it. Decide which image is the one users will flash via sysupgrade.

  11. DTS — pio node declared twice

  The DTS has two &pio { ... } blocks (one for spi0_flash_pins/i2c_pins, another for wf_dbdc_pins). Valid in DT, but normally merged. Cosmetic; also the pse_out_en/pse_int_n entries inside the first &pio block are
  written as if they were GPIO hog definitions but lack the gpio-hog property — they will be silently ignored.

  pse_out_en {
      gpios = <&pio 27 GPIO_ACTIVE_HIGH>;
      output-high;
  };

  Missing gpio-hog; and line-name. If this PSE enable is supposed to be driven high at boot, it never will be.

  12. scripts/mkits_sonicfi_rap630w_211g.sh — diverged copy of scripts/mkits.sh

  Almost verbatim copy of the upstream script. Only real additions: ROOTFS_NODE with crc32/sha1 hashes, the -r rootfs option handling. Carrying a 237-line fork for ~30 lines of behaviour is a maintenance burden.

  Fix: add the rootfs-hash feature to scripts/mkits.sh behind a flag, delete the fork.

  13. kernel/boot_param.c — no header, no users

  The 971-patch adds kernel/boot_param.c with 13 module parameters, all static except dual_boot. dual_boot is referenced via extern inside the same patch's init/do_mounts.c change. No header file is added, so
  anyone else who needs these variables has to extern them by hand. Also, the file has zero EXPORT_SYMBOLs — fine for built-in-only use, but worth noting.

  More concerning: the SPDX/license header says BSD-3-Clause, but the rest of the kernel file (and the Makefile line obj-y += boot_param.o) is GPL. Mixing BSD-3-Clause into a file linked into vmlinux is legally
  awkward. Normalise to GPL-2.0.

blogic:
also patches-25.12/0103-mediatek-add-support-for-Sonicfi-Rap630w-211g.patch needs to be split up into smaller patches that actually explain what they do. about 10% of the 0103 patch adds support for the board, the other 90% to various build system things,

maybe something like

	patches-25.12/0103-tools-add-crc32sum.patch
	patches-25.12/0104-kernel-add-boot-param-and-dual-image-rootfs.patch
	patches-25.12/0105-mediatek-mt7981b-add-hnat-dtsi-node.patch
	patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap630w-211g.patch

richardzhang-sw2:
Hi @blogic , thanks a lot for your advice. We have done improvement and made corrections, and updated the relevant files. But for item 7 about "02_network", it is correct design and configuration for rap630w-211g, no need to do modification. Please help check, thanks a lot!

richardzhang-sw2:
Hi @blogic, please help verify this PR and merge it. We would like to do the Allure auto test on the main image for rap630w-211g with kernel6.12. Thanks a lot!

blogic:
@richardzhang-sw2 there are now 4 patches stacked. please rebase upon latest main branch and just have a single unified commit in the PR to make it reviewable

richardzhang-sw2:
HI @blogic ,got it! We are doing a new PR#1066(https://github.com/Telecominfraproject/wlan-ap/pull/1066). Now the system is building FW images. Thanks a lot!

blogic:
duplicate, closing

PR #1059, opened 2026-04-22
WIFI-15455: Add sonicfi_rap63xc_211g support on openWRT25.12(Kernel6.12)

JackTsai-CBT:
Add sonicfi_rap63xc-211g configuration and functionality files on openWRT25.12 and Linux Kernel 6.12.


patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap63xc-211g.patch
```diff
@@ -0,0 +1,422 @@
...
+ . /lib/functions/ipv4.sh
++. /lib/functions.sh
+ 
+ [ -s $CFG ] || /bin/board_detect || exit 1
+ [ -s /etc/config/network -a -s /etc/config/system ] && exit 0
+@@ -310,6 +311,13 @@ generate_switch() {
+ }
+ 
+ generate_static_system() {
++	local LOG_SIZE="128"
```
blogic, line 35: custom hacks that are board specific in generic code, why do we need this ?

patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap63xc-211g.patch
```diff
@@ -0,0 +1,422 @@
...
++board=$(board_name)
+ 
+ OVERLAY="$( grep ' /overlay ' /proc/mounts )"
+ 
+@@ -24,7 +25,11 @@ released)
+ 	elif [ "$SEEN" -ge 5 -a -n "$OVERLAY" ]
+ 	then
+ 		echo "FACTORY RESET" > /dev/console
+-		factoryreset -y && reboot &
++        if [ "$board" = "sonicfi,rap63xc-211g" ]; then
```
blogic, line 71: custom hacks that are board specific in generic code, why do we need this ?

this just hides a bug where the board does nto reboot ?

patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap63xc-211g.patch
```diff
@@ -0,0 +1,422 @@
...
++++ b/package/base-files/files/etc/rc.local
+@@ -1,4 +1,23 @@
+ # Put your custom commands here that should be executed once
+ # the system init finished. By default this file does nothing.
+ 
++. /lib/functions.sh
++board=$(board_name)
++
++if [ "$board" = "sonicfi,rap63xc-211g" ]; then
++    if [ ! -f /sbin/reboot ] || [ -L /sbin/reboot ] || ! grep -q "reboot -f" /sbin/reboot 2>/dev/null; then
```
blogic, line 91: custom hacks that are board specific in generic code, why do we need this ?

blogic:
please remove all the custom hacks in the generic code.

patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap63xc-211g.patch
richardzhang-sw2: Hi @blogic ,  log_size  for logread is increased from 64 kB to 128 kB in openWRT25.12. But when testing rap63xc-211g and rap630w-211g images, the allure auto test was stuck in the configuration command:
./openwrt_ctl.py --prompt root@xxxxxxxxx -s serial --log stdout --user root --passwd openwifi -t /dev/ttyUSB0 --action cmd --value "logread".
If log_size was changed back to 64kB, the test could run normally.
So need to use 64kb for sonicfi projects. 
Thanks!

patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap63xc-211g.patch
richardzhang-sw2: Hi @blogic ,
Need to replace “reboot“ with "reboot -f" for rap630xc-211g. 
The reason is:
Under the MT7621 + MT7915 combination, the soft reboot freeze issue shares the exact root cause reported for the BPI-R4 (MT7988A) — Issue 3:After loading the mt7915_rom_patch.bin firmware, the Wi-Fi co-processor enters a special state, which prevents the SoC soft reboot sequence from completing. The system outputs "Restarting system" and then hangs indefinitely; power cycling is mandatory to restore operation.
Thanks a lot!

patches-25.12/0106-mediatek-add-support-for-Sonicfi-Rap63xc-211g.patch
richardzhang-sw2: Hi @blogic , the same reason as the previous comment, need to  replace “reboot“ with "reboot -f" for rap630xc-211g.
Thanks a lot!

blogic:
duplicate

PR #1064, opened 2026-04-23
After upgrading WiFi 5 APs to v2.11.2, using Insta server certificates not able to perform RTTY

abhivyas1992:
After upgrading WiFi 5 APs to v2.11.2, using Insta server certificates not able to perform RTTY


tanyasingh-ec:
Merged

PR #1065, opened 2026-04-24
WIFI-15459 WIFI-7 NAND flash boot failure on next branch

Justin5763:
qca-wifi-7: QCA UBI rootfs require the rootfs to be named "ubi_rootfs".
			this patch adds support for the rootfs name.


blogic:
merged, Thanks !

PR #1067, opened 2026-04-27
ucentral-schema: modify state.uc to query sub-interfaces

tanyasingh-ec:
Fixes: WIFI-15469


blogic:
merged into staging-for-v4.2.0-LTS branch, Thanks !

PR #1075, opened 2026-05-06
WIFI-15476 ATH SPF 13.1 CSU1 Upgrade: RTL8261N does not support 10G link speed for RAP7110C-341X

jemmy7780:
Fixed RTL8261N 10G mode issue: In RAP7110C-341x(IPQ53xx+RTL8261N),  when the switch port connects a 10G device, the link speed does not show "10G”.


blogic:
A number of things wrong with this submission:

  1. Misleading subject. The subject reads as a statement about silicon limitations ("RTL8261N does not support 10G link speed"), but the RTL8261N does support 10G. The actual problem is that
  rtl826xb_get_features() unconditionally clears ETHTOOL_LINK_MODE_10000baseT_Full_BIT for the RTL8261N PHY type. The subject should describe the fix, not the symptom, for example: qca-wifi-7: ipq53xx:
   rtk_phy: allow 10G link mode for RTL8261N.
  2. Empty commit message body. There is no explanation of where the bit is being cleared, why the existing switch is incorrect, or what the user-visible effect on RAP7110C-341X is. A reader has to
  reconstruct the bug from the one-line diff. Please include a body that explains the root cause and the fix.
  3. No subsystem prefix. Convention in this tree is <feed>: <board/subsystem>: <component>: short description. A simple WIFI-xxxxx followed by free-form text breaks that pattern and makes the log
  harder to scan.
  4. Malformed author and Signed-off-by. Both the From: and Signed-off-by: lines use the email address as the display name (From: "Yingwei.Huang@cybertan.com.tw" <Yingwei.Huang@cybertan.com.tw>).
  Please configure git config user.name to your real name, so commits show as Yingwei Huang <Yingwei.Huang@cybertan.com.tw>.
  5. Whitespace style violation. The surrounding switch block in rtl826xb_get_features() is indented with four spaces. The new case and break lines introduce a hard TAB, and the break mixes TAB plus
  four spaces. Match the existing style, eight spaces for case, twelve for break.
  6. Comment style. case RTK_PHYLIB_RTL8261N://support 10G modes is not consistent with the rest of the file (which uses /* ... */ block comments) and is missing the space after : and //. Either drop
  the comment, since a properly worded case label is self-explanatory, or use a normal block comment.
  7. Does not address the underlying mess. After this patch the switch in rtl826xb_get_features() reads as:

  switch (priv->phytype) {
      case RTK_PHYLIB_RTL8251L:
      case RTK_PHYLIB_RTL8254B:
          linkmode_clear_bit(...10000baseT...);
          break;
      case RTK_PHYLIB_RTL8261N:
          break;
      default:
          linkmode_clear_bit(...10000baseT...);
          break;
  }

  7. The RTL8251L/RTL8254B arm is now functionally identical to default, so it serves no purpose and is dead code. The proper fix is to invert the logic so the switch expresses its real intent, namely
  "clear 10G unless we are on a PHY that supports it":

  switch (priv->phytype) {
      case RTK_PHYLIB_RTL8261N:
          /* supports 10G */
          break;
      default:
          linkmode_clear_bit(...10000baseT...);
          break;
  }

Please respin the patch with: a fix-style subject with subsystem prefix, an explanatory body, a clean author/SoB, corrected whitespace, the dead RTL8251L/RTL8254B arm removed, and the Fixes: trailer
  pointing at the introducing commit (or Fixes: WIFI-15476 if that is how this is being tracked).

YingWei-cybertan:
@blogic I have made the requested changes and committed them to a new branch. Could you please review it again? Thank you.Commit URL:https://github.com/Telecominfraproject/wlan-ap/commit/5a936084fae19dfcb29e2f25a8b5230a40a9f4d6

blogic:
1. The commit message was passed via git commit -m "..." with embedded newlines, so the entire root cause / solution / Fixed: body got swallowed into the Subject: line. Running git log --oneline will
   print one ~400-character line. Ask them to re-shape so the subject is a single 50–72-char summary and the rest is body, separated by a blank line:

  Subject: qca-wifi-7: ipq53xx: rtk_phy: allow 10G for RTL8261N

  <blank line>
  root cause: rtl826xb_get_features() unconditionally clears
  ETHTOOL_LINK_MODE_10000baseT_Full_BIT for every Realtek PHY type,
  including the 10G-capable RTL8261N.

  solution: take the RTL8261N out of the default-clear path so it
  keeps the 10000baseT_Full link mode.

  Fixes: WIFI-15476 
  
  
  2. Indentation mismatch inside the switch. The surrounding function uses 8-space case-arm indent (Realtek house style); the new arms are 4-space. The diff context shows     switch (priv->phytype) (4
  leading spaces) then the old         case ... (8) is being replaced by     case ... (4). Leaves the function with mixed indent. Either re-indent the new arms to 8-space to match, or reflow the whole
  switch in one pass.

YingWei-cybertan:
@blogic  I have made the requested changes and committed them to a new branch. Could you please review it again? Thank you.

blogic:
merged, Thanks !

blogic:
turns out this patch enables 10G on all boards causing a regression, I will revert it and we need to figure out a better way to indicate if 10G should be enabled or now.

jemmy7780:
Hi sir.
I see that the patch in #1085 is based on our patch. If you remove our patch, their patch will fail to apply. Moreover, the current approach remains consistent with the v4.2.X practice, i.e., by default RTL8261 supports 10G. For projects that do not support 10G, like WF189, we handle it by adding 'limit_rtlphy_10g_ability' in the DTS file and then impose corresponding restrictions in the rtk driver.
so I don't think we need revert it.  please give us your suggestion, thanks.

blogic:
we will revert this patch and not merge the regression fix from CIG. cybertan needs to provide an opt-in verion of their patch. weesntially the reverse logic of #1085 (which we wont merge)

jemmy7780:
OK, we will do it.

jemmy7780:
I'm still a bit confused: RTL8261 supporting 10G speed should be a common requirement. This can be seen from the old TIP v4.2.X, where 10G was supported by default, and only the WF189 project was special and needed a restriction, so CIG issued a patch to fix it. Back to TIP 5.0, the RTK driver code lacks RTL8261 10G speed support by default, and our patch is exactly fixing this issue. Now, if we follow your request, then for any other project that also needs RTL8261 10G speed support, they would have to make and add our patch again.
----------------------- TIP v4.2.X --------------------------------
static int rtl826xb_get_features(struct phy_device *phydev)
{
    int ret;
    struct device_node *np;
    ret = genphy_c45_pma_read_abilities(phydev);
    if (ret)
        return ret;

    linkmode_or(phydev->supported, phydev->supported, PHY_BASIC_FEATURES);


    linkmode_set_bit(ETHTOOL_LINK_MODE_2500baseT_Full_BIT,
                       phydev->supported);
    linkmode_set_bit(ETHTOOL_LINK_MODE_5000baseT_Full_BIT,
                       phydev->supported);

    /* not support 10M modes */
    linkmode_clear_bit(ETHTOOL_LINK_MODE_10baseT_Half_BIT,
                       phydev->supported);
    linkmode_clear_bit(ETHTOOL_LINK_MODE_10baseT_Full_BIT,
                       phydev->supported);

    **np = of_find_node_by_name(NULL, "mdio");
    if (np)
        if (of_property_read_bool(np, "limit_rtlphy_10g_ablity"))
        {
                linkmode_clear_bit(ETHTOOL_LINK_MODE_10000baseT_Full_BIT, phydev->supported);
        }**


    return 0;
}

blogic:
adding your patch breaks other boards. so we revert it. you need to provide a way to enable 10G on your board without breaking all other 1g and 2.5g consumers of the driver

YingWei-cybertan:
@blogic Hello blogic,We have just submitted a new pull request for this issue. Thank you!
https://github.com/Telecominfraproject/wlan-ap/commit/3e1eb32c05b9bd89bd945d6e46b4cef88ec54f98

blogic:
merged, Thanks !

PR #1074, opened 2026-05-06
ramoops dts for udaya a6-id2

abhivyas1992:
Udaya_A6_ID2: Crash logs not getting saved in Pstore

https://telecominfraproject.atlassian.net/browse/WIFI-15423

PR #1076, opened 2026-05-08
Wifi 15423 wifi6 udaya a6 ramoops crashlog

abhivyas1992:
Udaya_A6_ID2: Crash logs not getting saved in Pstore

https://telecominfraproject.atlassian.net/browse/WIFI-15423

Also PR for release/v4.2.0 branch : 
https://github.com/Telecominfraproject/wlan-ap/pull/1074


blogic:
merged, Thanks !

1. Wrong mechanism. The change should be applied directly to the DTS file in the tree, not introduced as an additional patch file under patches/. We carry the qca-wifi-6 IPQ50xx DTS sources under
  feeds/qca-wifi-6/ipq50xx/files/, so the correct approach is to modify qcom-ipq5018-udaya-a6-id2.dts in place and commit that change. The patches/ subdirectories are reserved for fixups against
  upstream code we cannot edit directly, which is not the case here.
  2. Wrong path. The patch targets feeds/ipq807x_v5.4/ipq50xx/patches/, which does not exist in this tree. The IPQ50xx sources live in feeds/qca-wifi-6/ipq50xx/. Even if the "add a patch file" approach
   were acceptable, this drop-in would have had no effect because the directory it references is not part of our build.
  3. Patch format. The embedded diff appears to have been produced with plain diff -u original.orig original, leaving .orig suffixes and local timestamps in the header lines. Patches in this tree are
  expected to come from git format-patch for tree changes, or from quilt refresh for the rare cases that genuinely belong in a patches/ directory. Neither produces this style.
  4. Commit message hygiene. The subject ramoops dts for udaya a6-id2 lacks a subsystem prefix, omits the WIFI-xxxxx tracking ID, and there is no body explaining the motivation or the affected memory
  layouts. Please include the appropriate Fixes: trailer (in this case Fixes: WIFI-15423) in future submissions.

PR #1077, opened 2026-05-09
Fixes for issues in ATH13.1

psvtrajan:
**Changes:**
	wifi-scripts:
		- Generating different macaddr for interfaces
		- Updating proper value to wpa in hostapd configuration when security mode is configured as psk
		- Updating wpa_psk_file and vlan_file in the hostapd configuration when multi-psk is configured
	hostapd:
		- Copying mpskd to the rootfs

**Fixes:** WIFI-15425, WIFI-15427, WIFI-15435, WIFI-15437, WIFI-15441


blogic:
updated patch annotation and manually figured out all the correct Fixes: tags myself. please ensure this is done properly for future patches.

merged, Thanks !

PR #1079, opened 2026-05-11
qca-wifi-7: revise Asterfusion AP7330 to the new ATH13.1 (WIFI-15482)

Yang-Yongzhi:
For new ATH13.1 and kernel:
- Enable the CONFIG_GPIO_WATCHDOG of kernel config 6.6 (This option was previously enabled in the 6.1 config but was missing from the newly created 6.6 configuration files, causing the existing DTS watchdog node to fail to bind silently.)
- sync DTS to newer SDK and kernel version.

FIX:
- The Reset button is not work because of wrong gpio number and wrong button function.

Fixes: WIFI-15482


blogic:
The commit message claims "enable CONFIG_GPIO_WATCHDOG" but omits thre important details: the DTS already declares the gpio-watchdog node, and the driver has been silently failing to bind because the kernel option was never enabled. It also fails to mention that the DTS is being synced to a newer SDK/kernel version, and that one of the buttons was previously broken and is being corrected here.

Finally, the ticket ID belongs on a Fixes: WIFI-15482 trailer immediately above the Signed-off-by line, not in the subject.

Yang-Yongzhi:
@blogic 
Thank you for the detailed review and the catch on those missing context.
I have updated the commit message.
Please take another look when you have a chance. 
Thanks!

blogic:
merged, Thanks !

PR #1083, opened 2026-05-12
rrmd: Add 5 sec delay before checking channel switch status

tanyasingh-ec:
Fixes: WIFI-15488

Add a 5-second delay before checking channel switch status. hostapd may still be processing the switch even on non-DFS/non-weather channels, causing the script to incorrectly report a failed channel switch.


blogic:
merged, Thanks!

PR #1084, opened 2026-05-13
WIFI-15485-CIG-WF-189H-wifi-board-id-issue

Cyrusx7:
CIG WF-189H Due to radio issues caused by the board ID in DTS used in 189h, the board ID has been corrected


blogic:
merged, Thanks !

PR #1085, opened 2026-05-13
WIFI-15489-WF-189-add-speed-limit-condition-to-rtl8261n-for-WF-189-WAN-port

Justin5763:
WF-189 WAN port does not use 10G mode, we add limit_rtlphy_10g_ablity to DTS and check it in RTL code.

### DIFF
--- a/feeds/qca-wifi-7/ipq53xx/files-6.6/drivers/net/phy/rtk/rtk_phy.c
+++ b/feeds/qca-wifi-7/ipq53xx/files-6.6/drivers/net/phy/rtk/rtk_phy.c
@@ -5,6 +5,7 @@
  */
 
 #include <linux/module.h>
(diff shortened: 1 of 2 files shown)


blogic:
I would like to turn this around logic wise. the patch introducing the 10G support clearly causes a regression on you board and potentially others. the 10G should be opt-in and not opt-out

Justin5763:
https://github.com/Telecominfraproject/wlan-ap/pull/1075 
I found this commit enables 10G support for all boards. If this functionality should be opt-in, maybe ask the author to limit 10G support to their board only.I don’t think it’s appropriate for me to change their code arbitrarily.

blogic:
@Justin5763 correct I will revert #1075 and close this PR. Thanks for catching this regression

PR #1087, opened 2026-05-13
Disable energy-detect CCA on HaLow interface bring-up

ian77chen:
Disable energy-detect CCA on HaLow interface bring-up by
invoking 'morsectrl edconfig energy ignore' in drv_morse_setup.

new package "morsectrl"

Fixes: WIFI-15483


blogic:
merged, Thanks !

PR #1088, opened 2026-05-13
qca-wifi-7: fix WAN/LAN port assignment for AP7330

Yang-Yongzhi:
Change eth0 from WAN to LAN interface, and reassign switch port 4 as the WAN port.

Fixes: WIFI-15490


blogic:
merged, Thanks!

PR #1090, opened 2026-05-14
ath12k: pass the correct value of each TID during a stop AMPDU session

ArifAlam:
backport:
https://lore.kernel.org/all/20260227110123.3726354-1-reshma.rajkumar@oss.qualcomm.com/

With traffic ongoing for data TID [TID 0], an DELBA request to stop AMPDU for the BA session was received on management TID [TID 4]. The corresponding TID number was incorrectly passed to stop the BA session, resulting in the BA session for data TIDs being stopped and the BA size being reduced to 1, causing an overall dip in TCP throughput.

Fix this issue by passing the correct argument from ath12k_dp_rx_ampdu_stop() to ath12k_dp_arch_peer_rx_tid_reo_update() during an AMPDU stop session. Instead of passing peer->dp_peer->rx_tid, which is the base address of the array, corresponding to TID 0, pass the value of &peer->dp_peer->rx_tid[params->tid]. With this, the different TID numbers are accounted for.


ArifAlam:
Will resubmit the patch in mbox format which will retain upstream detail / description.

blogic:
@ArifAlam is this confirmed to fix the issue ? would merge it into the main branch in that case

blogic:
merged, Thanks !

PR #1095, opened 2026-05-19
qca-wifi-7: update BDF (5G and 6G radios) for Edgecore EAP105 based on SPF12.5

tanyasingh-ec:
Fixes: WIFI-15436


blogic:
merged, Thanks !

PR #1096, opened 2026-05-19
mediatek: mac80211: decrease passive scan channel time

mkwacz:
Compensate for FW channel switch and wiphy_delayed_work scheduling overhead in kernel 6.x by subtracting a fixed overhead from passive scan dwell time.

Fixes: WIFI-14822


blogic:
merged, Thanks !

PR #1098, opened 2026-05-21
radio.uc: Skip match_htmode() for HaLow band to avoid warning reject

ian77chen:
radio.uc: Skip match_htmode() for HaLow band to avoid warning reject
("Selected radio does not support requested HT mode HE80, falling back to VHT160")

Fixes: WIFI-15497


blogic:
merged, Thanks !

PR #1107, opened 2026-05-27
WIFI-15508-CIG-WF-189-The-OpenWiFi-version-based-on-ATH13.1CSU1-current-latest-version-has-a-low-TP-testing-rate

Cyrusx7:
…rrent latest version) has a low TP testing rate

Set the default configuration in config_generate to '0' for network.globals.packet_steering. Previously, packet_steering was enabled by default, which resulted in uneven CPU load during TP testing


blogic:
this will have an impact on all other non QCA boards.

Kenshi0000:
John, if that, shall we only modify it for CIG products or all QCA platform based ?

PR #1108, opened 2026-05-27
ucentral-schema: fix HaLow center frequency shows wrong in state.uc

ian77chen:
Fixes: WIFI-15501

Root Cause: 
The state reporting code calculates HaLow frequency incorrectly. It computes lower and upper edge frequencies using center_freq minus/plus bandwidth/2, then reports those two edge values as the frequency array. This does not match the expected format where channels and frequencies should be 1-to-1 mapped, same as 2.4G and 5G radios. The survey frequency variable also uses this incorrect center_freq value.

Solution: 
Patch 004-fix-halow-freq-report.patch. Replace the edge frequency calculation with a per-channel frequency lookup from the S1G mapping table. Build both s1g_channels and s1g_frequencies arrays in parallel, so each channel maps to its correct S1G center frequency. Update the survey frequency to use the mapped S1G frequency directly.

### DIFF
--- a/feeds/ucentral/ucentral-schema/patches/004-fix-halow-freq-report.patch
+++ b/feeds/ucentral/ucentral-schema/patches/004-fix-halow-freq-report.patch
@@ -1,0 +1,48 @@
+--- a/system/state.uc
++++ b/system/state.uc
+@@ -524,36 +524,31 @@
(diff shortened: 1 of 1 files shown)


blogic:
merged, Thanks !

PR #1109, opened 2026-05-27
ucentral-schema: fix radio 2.4G/5G shows empty when HaLow radio exists

ian77chen:
Fixes: WIFI-15504

Root Cause: 
State reporting skips NOHT interfaces at ucentral-schema level: Once radios are in NOHT/DISABLED state with wiphy_freq=0, the iface.uc module skips them entirely. Then state.uc cannot find the interface data and skips the entire radio from the state report. Cloud only sees the HaLow radio which is functioning normally.

Solution: Patch 005-fix-radio-noht-skip.patch. Remove the wiphy_freq=0 filter in iface.uc so interfaces with unknown channel are still reported. Modify state.uc to gracefully handle empty frequency arrays instead of skipping the radio.


blogic:
merged, Thanks !

PR #1113, opened 2026-05-28
qca-wifi-7: hostapd: preserve HT/VHT/HE/EHT capabilities before channel switch

SebastianHuang-ec:
ref.: https://www.spinics.net/lists/hostap/msg11543.html

The HT and VHT capability information are missing after the CSA and CAC timeout. Even if the AP is run with 801.11ax or 802.11be mode, some of the clients would ignore the SSID when there is no HT and VHT IEs. 

The patch explicitly preserve current HT/VHT/HE/EHT capabilities before actual channel switch takes place, so the information and be used after CSA.

Fixes: WIFI-15499


blogic:
do we also need this in the main branch ? LTS backports require a merge into the main branch prior.

SebastianHuang-ec:
I cannot reproduce the issue on main branch firmware. That's the reason that I only send PR for v4.2 LTS.

blogic:
merged, Thanks !

PR #1114, opened 2026-05-28
rrmd: accept per-BSS hostapd ubus objects with dotted ifnames

mkwacz:
On ath13.1 multi-BSS, hostapd publishes one ubus object per BSS as hostapd.<phy>.<radio>-<bss>, e.g. hostapd.phy00.1-0. The split(.) based parser in interfaces_subunsub required exactly two dot-separated components and rejected those names, so the interfaces[] table stayed empty and every BSS-targeted RRM action (channel_switch, tx_power, kick by addr->bssid) returned "failed to trigger" or "station unknown".

Replace the 2-part split with a "hostapd." prefix strip so dotted ifnames are preserved while remaining backward compatible with hostapd.wlan0 style names on older platforms.

Fixes: WIFI-15434


blogic:
merged, Thanks !

PR #1117, opened 2026-06-03
WIFI-15508-CIG-WF-189-The-OpenWiFi-version-based-on-ATH13.1CSU1-current-latest-version-has-a-low-TP-testing-rate-NEW

Cyrusx7:
…rrent latest version) has a low TP testing rate

	Set the default value of network.globals.packet_steering to '0'
for QCA-based products. Previously, packet_steering was enabled by default, which caused uneven CPU load distribution across cores during throughput testing.

	This change is limited to QCA platform targets only, as the
packet steering behavior and its impact are specific to QCA chipset architecture (e.g., IPQ5332, IPQ5425).


blogic:
merged, Thanks !

PR #1118, opened 2026-06-04
rrmd: Add 5 sec delay before checking channel switch status

tanyasingh-ec:
- Fix typo: change sleep time from 5ms to 5000ms

Fixes: WIFI-15488


blogic:
merged, Thanks !

PR #1122, opened 2026-06-05
WIFI-15511-DFS-is-broken-on-WF196-APs-on-5.0.0

Cyrusx7:
Following the previous approach, the 6G country code is not
 configured by default, which also avoids firmware issues


blogic:
merged, Thanks !

PR #1123, opened 2026-06-09
WIFI-15451:fix 5GHz radio fails to initialize with EHT config allow-dfs false instead of falling back to HE

hehuimin-cybertan:
fixed WIFI-15451:fix 5GHz radio fails to initialize with EHT config allow-dfs false instead of falling back to HE


blogic:
I made some cleanups, pushed the patch to the proper repository and updated the ucentral-schema package.

merged a modified version of the patch, Thanks !

PR #1124, opened 2026-06-09
qca-wifi-7/wifi-scripts: fix wpa_supplicant fails to start up sometimes

jemmy7780:
Enclosing macaddr_base in double quotes can resolve the parsing issue.

fixes: WIFI-15524


blogic:
merged, Thanks !

PR #1125, opened 2026-06-09
WIFI-15442: Fix Clients fail to obtain IP in NAT mode at TIP-devel-168e840 on Cybertan_rap630c-311g

YingWei-cybertan:
Fixes a netifd module issue:
After the bridge clears all members when handling DEV_EVENT_REMOVE, the bridge fails to wake up again when a new member is added later.It usually occurs inside a bridge consisting entirely of Wi-Fi interfaces, because in this case it triggers a condition where all bridge members are cleared and then added again.
https://telecominfraproject.atlassian.net/browse/WIFI-15442


blogic:
I have just pushed a fix to main for WIFI-15468, I have a feeling that this might also fix this issue without the patch in this PR needing to be applied, please test current main branch and let me know if it fixes the problem

jemmy7780:
hi @blogic 
      I just verified your patch, and it does not resolve our issue.

blogic:
I have just attached a different approach as a patch to the jira ticket, could you help testing if that fixes the issue when applied to the main branch ?

YingWei-cybertan:
Submit  a new PR at https://github.com/Telecominfraproject/wlan-ucentral-schema/pull/98 to resolve this issue.

blogic:
closing, this was fixed using a schema change instead

PR #1129, opened 2026-06-11
hostapd/ucentral-state: fix reload crash on dead radio and add radio …

tanyasingh-ec:
…self-healing

hostapd: guard iface.state() call with type check in iface_reload_config(). When a radio disappears (ENODEV), the native interface object may remain in hostapd.interfaces but with an invalid .state property. This caused "left-hand side is not a function" errors on every config reload attempt. The function already returns false for non-ENABLED interfaces, so the type guard simply takes the same path without throwing.

ucentral-state: extend self_healing() to check health_stat.data.radios for per-radio SSID failures. When any radio reports failed_ssids, trigger a network restart (subject to the existing 5-minute cooldown). Added diagnostic logging for both interface and radio triggered restarts.

Fixes: WIFI-15549

### DIFF
--- a/feeds/qca-wifi-6/hostapd/files/hostapd.uc
+++ b/feeds/qca-wifi-6/hostapd/files/hostapd.uc
@@ -538,7 +538,7 @@ function iface_reload_config(name, phydev, config, old_config)
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


blogic:
merged, Thanks !

PR #1130, opened 2026-06-11
ucentral-schema: add per-radio health check and self-healing support

tanyasingh-ec:
The global SSID lookup in health.uc could not detect a failed radio on multi-band APs because the same SSID running on a sibling radio masked the failure. This patch adds per-radio validation by matching SSIDs to their expected phy index or ifname_prefix.

Additonally, support path+N device node notation for multi-radio chips that expose multiple phys under a single sysfs path, and add phy_band matching for reconf-capable devices where multiple wifi-devices share the same phy but operate on different bands.

Changes:
- renderer/wifi/iface.uc: expose phy index and has_channel flag for each interface so dead-radio VAPs remain visible to health checks 
- system/health.uc: add check_radio_health() that validates each non-disabled radio has its expected SSIDs active with a channel assigned; failures are reported in state.radios and factored into the sanity score

Fixes: WIFI-15549


blogic:
merged, Thanks !

PR #1132, opened 2026-06-12
qca-wifi-7: hostapd: restore global ban and seconds ban_time in del_c…

mkwacz:
…lient

The ath13.1.5 SDK update overwrote the hostapd ubus overlay src/ap/ubus.c with the vendor version and dropped two ban features:

- ban_time was armed with eloop_register_timeout(0, time * 1000, ...), so the value was treated as microseconds and an N second ban expired after N ms.
- global_ban (DEL_CLIENT_GLOBAL_BAN) and the per-interface fan-out were removed, so a kick only banned the single BSS and the station immediately reassociated on the other band.

Restore the seconds based timeout and the global_ban handling, reinstating the behaviour originally added in 56e0b3b50cf4 ("hostapd: globally ban clients when they are kicked"), so a kicked station is banned across all BSSes for the requested duration.

Fixes: 45d2d0b129c3 ("qca-wifi-7: update feed to ath13.1.5 SDK")
Fixes: WIFI-15434

### DIFF
--- a/feeds/qca-wifi-7/hostapd/wlan-app-extns/hostapd/patches/zzz-t05-hostapd-ubus-global-ban-and-seconds.patch
+++ b/feeds/qca-wifi-7/hostapd/wlan-app-extns/hostapd/patches/zzz-t05-hostapd-ubus-global-ban-and-seconds.patch
@@ -1,0 +1,85 @@
+diff --git a/src/ap/ubus.c b/src/ap/ubus.c
+--- a/src/ap/ubus.c
++++ b/src/ap/ubus.c
+@@ -137,7 +137,7 @@
+ 		}
+ 	}
+ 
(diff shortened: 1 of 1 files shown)


psvtrajan:
@mkwacz @blogic 
Suggestion:
Make this change as a patch, that gets applied while compiling. This will avoid this kind of issue in future while upgrading to the latest SDK.

blogic:
I fully agree, please generate a patch for this fix rather than patching ubus.c directly

mkwacz:
my bad, will update it.

mkwacz:
Patch ready for review, validated in https://telecominfraproject.atlassian.net/browse/WIFI-15434

blogic:
merged, Thanks !

PR #1133, opened 2026-06-12
qca-wifi-7: ath12k-wifi: update FCC ath13.1 BDF for RAP750e-h and RAP…

jemmy7780:
…750e-s

Update CTL power and target power for FCC BDF

Fixes: WIFI-15564


blogic:
Merged, Thanks !

PR #1135, opened 2026-06-15
WIFI-15548 CA BDF not present for WF189* models

Cyrusx7:
Add CA BDF to support CA country code for v4.2.5


blogic:
merged, Thanks !

PR #1136, opened 2026-06-15
patches-25.12: fix EAP115/EAP115a Ethernet ports issues

SebastianHuang-ec:
Make mt798x internal 2.5GbE port driver to be loaded earlier. 
Also adjust device tree to include required entries for both ports.

Fixed: WIFI-15477


blogic:
merged, Thanks !

PR #1137, opened 2026-06-17
WIFI-15478:Openwrt-25.12 AP selects non PSC channel after ACS even with acs-exclude-6ghz-non-psc parameter enabled

hehuimin-cybertan:
fixed WIFI-15478:Openwrt-25.12 AP selects non PSC channel after ACS even with acs-exclude-6ghz-non-psc parameter enabled


blogic:
I made some cleanups to the patch.

Merged, Thanks !

PR #1138, opened 2026-06-24
mediatek: mt7981b-emplus-wap588m: set in-band-status on LAN GMAC

cpchangemplu:
The LAN port (gmac0, SGMII via phy1) had no managed link mode, so
phylink never took link state from the SGMII PCS and the port could
not transmit. Set managed = "in-band-status" so link/speed/duplex come
from SGMII in-band signalling, bringing the LAN port up.

Tested on WAP588M: LAN links up and passes traffic.

Fixes: WIFI-15572

### DIFF
--- a/feeds/mediatek/mediatek/dts/mt7981b-emplus-wap588m.dts
+++ b/feeds/mediatek/mediatek/dts/mt7981b-emplus-wap588m.dts
@@ -187,47 +187,42 @@
 	};
 };
 
-&eth {
-        status = "okay";
+&mdio_bus {
+	phy1: ethernet-phy@1 {
+		compatible = "ethernet-phy-id001c.c916";
+		reg = <1>;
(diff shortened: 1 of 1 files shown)


cpchangemplu:
https://telecominfraproject.atlassian.net/browse/WIFI-15572

blogic:
patches need to be in main before they can be merged into the LTS branch

cpchangemplu:
> patches need to be in main before they can be merged into the LTS branch

Change base to release/v4.2.0

tanyasingh-ec:
Please update the PR and the commit msg as per the guidelines.

tanyasingh-ec (changes requested):
Please update the PR and the commit msg as per the guidelines.

mkwacz:
@cpchangemplu Please also comment if we need this change in current main. Thanks.

PR #1139, opened 2026-06-24
mediatek: mt7981b-emplus-wap588m: exchange lan port and wan port

cpchangemplu:
Per the customer requirement, LAN port need to be eth1 , WAN port need to be eth0

Tested on WAP588M: eth0 and eth1 have been inter-exchanged

Fixes: WIFI-15578

### DIFF
--- a/feeds/mediatek/mediatek/dts/mt7981b-emplus-wap588m.dts
+++ b/feeds/mediatek/mediatek/dts/mt7981b-emplus-wap588m.dts
@@ -187,47 +187,42 @@
 	};
 };
 
-&eth {
-        status = "okay";
+&mdio_bus {
+	phy1: ethernet-phy@1 {
+		compatible = "ethernet-phy-id001c.c916";
+		reg = <1>;
+		phy-mode = "sgmii";
+		reset-gpios = <&pio 39 GPIO_ACTIVE_LOW>;
+		reset-assert-us = <120000>;
+		reset-deassert-us = <120000>;
+	};
+};
 
-        gmac0: mac@0 {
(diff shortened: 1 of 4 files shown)


blogic:
patches need to be in main before they can be merged into the LTS branch

cpchangemplu:
> patches need to be in main before they can be merged into the LTS branch

Change base to release/v4.2.0

mkwacz:
I think that in the first step you should merge it into the main, then cherry-pick into staging-for-v4.2.0-LTS

venkatchimata:
@cpchangemplu is this already in main? If not, could you please open a pr for main and then to the release branches?

mkwacz:
@venkatchimata I've checked current main and we don't have patches /feeds/mediatek/ on main. It means that we need to add the same logic for main branch by patches-25-12 to avoid the same issues in the feature.
Now I think that base of change is correct but @cpchangemplu should create a new PR for main with the same change if needed.

mkwacz:
@cpchangemplu Please also update commit message as per the guidelines.

mkwacz (changes requested):
Fix commit msg as per the guidelines.

tanyasingh-ec:
> @venkatchimata I've checked current main and we don't have patches /feeds/mediatek/ on main. It means that we need to add the same logic for main branch by patches-25-12 to avoid the same issues in the feature. Now I think that base of change is correct but @cpchangemplu should create a new PR for main with the same change if needed.

@cpchangemplu for 5.1 MTK, we're using https://github.com/Telecominfraproject/wlan-ap/tree/staging-mtk-v5.0 for development

PR #1140, opened 2026-06-24
feeds: mt76: add golden board data file for Edgecore EAP115

SebastianHuang-ec:
Read board data from the golden file instead of from MTD partition.

Fixed: WIFI-15575


blogic:
merged, Thanks !

PR #1142, opened 2026-06-25
Staging wifi 15574 udaya a5 id2 capabilities mac support

abhivyas1992:
https://telecominfraproject.atlassian.net/browse/WIFI-15574

### DIFF
--- a/feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
+++ b/feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
@@ -15,7 +15,8 @@ reload_service() {
 }
 
 start_service() {
-	[ -f /etc/ucentral/capabilities.json ] || /usr/share/ucentral/capabilities.uc
+	[ -s /etc/ucentral/capabilities.json ] && rm -f /etc/ucentral/capabilities.json
+	/usr/share/ucentral/capabilities.uc
 	
 	[ -f /tmp/ucentral.version ] || cat /etc/openwrt_release | grep DISTRIB_REVISION= | cut -d\' -f2 > /tmp/ucentral.version
 
--- a/feeds/ucentral/ucentral-schema/patches/002-update-capabilities-udaya-a5-id2.patch
+++ b/feeds/ucentral/ucentral-schema/patches/002-update-capabilities-udaya-a5-id2.patch
@@ -1,0 +1,65 @@
+--- a/system/capabilities.uc
++++ b/system/capabilities.uc
+@@ -29,6 +29,60 @@
+ capa.compatible = replace(board.model.id, ',', '_');
+ capa.model = board.model.name;
+
++if (board.model.id == 'udaya,a5-id2') {
++	capa['default-config'] = true;
++
++	let default_config = {
++		country: 'US'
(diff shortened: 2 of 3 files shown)


feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
```diff
@@ -15,7 +15,8 @@ reload_service() {
 }
 
 start_service() {
-	[ -f /etc/ucentral/capabilities.json ] || /usr/share/ucentral/capabilities.uc
+	[ -s /etc/ucentral/capabilities.json ] && rm -f /etc/ucentral/capabilities.json
+	/usr/share/ucentral/capabilities.uc
```
blogic, line 19: why ?

sujeet765: `[ -f /etc/ucentral/capabilities.json ] || /usr/share/ucentral/capabilities.uc`
: if capabilities.json already exists, -f is true, || short-circuits, capabilities.uc is never run (skipped).

```
[ -s /etc/ucentral/capabilities.json ] && rm -f /etc/ucentral/capabilities.json
/usr/share/ucentral/capabilities.uc
```
: delete the file if it exists and is non-empty. this always regenerates the latest capabilities.json every time, regardless of whether one already existed.

phwhite: With lots of reboots it can cause unintential wear on the flash and reduce life time.    Also, why would capabilites need to be regenerated?  can you share some specific use-cases?

sujeet765: If we change in the radio configuration (e.g. country code), the capabilities do not update. The capabilities will not update unless we manually remove or delete `/etc/ucentral/capabilities.json`.

venkatchimata: @sujeet765 By change in the radio configuration, I am assuming the AP config file (json) that we apply through ucentral. If you change the country code in the config file it does n't impact /etc/ucentral/capabilities.json . Even if you regenerate the capabilities file after the config file, it does n't change.

Or if my above assumption of config file is not correct, please share the detailed steps of a scenario on why do we need to regenerate the capabilities file.

venkatchimata: I just saw your other  commits.
1. It looks like you made one patch for ucentral-schema ie feeds/ucentral/ucentral-schema/patches/002-update-capabilities-udaya-a5-id2.patch . This patch applies to all the platforms , not just Udaya-A5
2. This commit https://github.com/Telecominfraproject/wlan-ap/pull/1142/changes/db481d82e7c6de19c6e303fea863c827deb1d18c and https://github.com/Telecominfraproject/wlan-ap/pull/1142/changes/c8f4fbb5346d617b541a52109bd03498e12667ee are on the same patch. Please squash them to one.
3. There is one Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com> . Probably it is not needed?

PR #1143, opened 2026-06-25
Staging wifi 15563 fix udaya a6 id2 captive portal client removal

abhivyas1992:
https://telecominfraproject.atlassian.net/browse/WIFI-15563


venkatchimata:
@abhivyas1992 I think this commit needs proper rebasing.

PR #1144, opened 2026-06-26
hostapd: fix unrecognized osen and operator_icon

hehuimin-cybertan:
Problem
In the config delivered under this test case, the radius host and anqp_3gpp_cell_net formats are incorrect. In addition, osen and operator_icon are not recognized in the hostapd updated to ATH13.1.

Fix
Modify the radius host and anqp_3gpp_cell_net in the config to correct values. Additionally, since hostapd does not support osen and operator_icon, these two configuration items should be removed from hostapd.sh.

Verification
With these modifications, iwinfo can correctly display wireless information, and wireless functionality operates normally.

### DIFF
--- a/feeds/qca-wifi-7/wifi-scripts/files/lib/netifd/hostapd.sh
+++ b/feeds/qca-wifi-7/wifi-scripts/files/lib/netifd/hostapd.sh
@@ -1305,14 +1305,14 @@ hostapd_set_bss_options() {
 
 	set_default hs20 0
 	set_default disable_dgaf "$hs20"
-	set_default osen 0
+	#set_default osen 0
 	set_default anqp_domain_id 0
 	set_default hs20_deauth_req_timeout 60
 	if [ "$hs20" = "1" ]; then
 		append bss_conf "hs20=1" "$N"
 		append_hs20_icons
 		append bss_conf "disable_dgaf=$disable_dgaf" "$N"
-		append bss_conf "osen=$osen" "$N"
+		[ -n "$osen" ] && append bss_conf "osen=$osen" "$N"
 		append bss_conf "anqp_domain_id=$anqp_domain_id" "$N"
 		append bss_conf "hs20_deauth_req_timeout=$hs20_deauth_req_timeout" "$N"
 		[ -n "$osu_ssid" ] && append bss_conf "osu_ssid=$osu_ssid" "$N"
@@ -1324,7 +1324,7 @@ hostapd_set_bss_options() {
 		json_for_each_item append_hs20_oper_friendly_name hs20_oper_friendly_name
 		json_for_each_item append_hs20_conn_capab hs20_conn_capab
 		json_for_each_item append_osu_provider osu_provider
-		json_for_each_item append_operator_icon operator_icon
+		[ -n "$operator_icon" ] && json_for_each_item append_operator_icon operator_icon
 	fi
 
 	if [ "$eap_server" = "1" ]; then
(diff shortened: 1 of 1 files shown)


venkatchimata (approved):
@hehuimin-cybertan Please add some description for the commit on what is the problem and why your commit fixes the issue.

Please also change the first line of the commit to start with all lowercase. Detailed commit guidelines are coming soon. For now, you can make it like this

iwinfo: fix iwinfo for passoint R1 feature

What is the problem..  Please expand this
Why you commit fixes it. Please expand this
Please share the test results. Please expand this.

Fixes: WIFI-15525

Signed-off-by: huimin.he <huimin.he@cybertan.com.tw>

mkwacz:
I did some investigation and found that osen and operator_icon are not supported in hostap ath13.1.5. It makes sense to remove these options from hostapd.sh or add support.

feeds/qca-wifi-7/wifi-scripts/files/lib/netifd/hostapd.sh
```diff
@@ -1324,7 +1324,7 @@ hostapd_set_bss_options() {
 		json_for_each_item append_hs20_oper_friendly_name hs20_oper_friendly_name
 		json_for_each_item append_hs20_conn_capab hs20_conn_capab
 		json_for_each_item append_osu_provider osu_provider
-		json_for_each_item append_operator_icon operator_icon
+		[ -n "$operator_icon" ] && json_for_each_item append_operator_icon operator_icon
```
mkwacz: How did you validate that such change is needed if currently operator_icon is not supported? It looks really strange...

mkwacz:
fixes:WIFI-15525 instead of Fixes: WIFI-15525

feeds/qca-wifi-7/wifi-scripts/files/lib/netifd/hostapd.sh
hehuimin-cybertan: Yes, the hostapd in ATH13.1 does not support osen and operator_icon. The configuration pushed by the controller also does not include these two items, and ucentral only handles osen. So theoretically, these two config values are empty. However, hostapd.sh always generates these two config entries regardless of whether they have values, which causes parsing errors. Therefore, I added conditional checks in hostapd.sh to control their generation.
Since hostapd does not support these two settings, I think they can be removed directly

feeds/qca-wifi-7/wifi-scripts/files/lib/netifd/hostapd.sh
```diff
@@ -1305,14 +1305,14 @@ hostapd_set_bss_options() {
...
-	set_default osen 0
+	#set_default osen 0
 	set_default anqp_domain_id 0
 	set_default hs20_deauth_req_timeout 60
 	if [ "$hs20" = "1" ]; then
 		append bss_conf "hs20=1" "$N"
 		append_hs20_icons
 		append bss_conf "disable_dgaf=$disable_dgaf" "$N"
-		append bss_conf "osen=$osen" "$N"
+		[ -n "$osen" ] && append bss_conf "osen=$osen" "$N"
```
mkwacz: What do you think about removing osen and operator_icon?

hehuimin-cybertan: Since hostapd does not support osen and operator_icon, I think they should be removed. I will revise this PR later.

mkwacz (approved):

venkatchimata (approved):

PR #1150, opened 2026-06-30
fix: ucentral-client: SAN handling.

ec-sam:
Let `ucentral-client` matched the configured server hostname against not only the certificate’s Common Name (**CN**) but Subject Alternative Name (**SAN**) DNS entries.

Fixes: WIFI-15479


mkwacz:
Patch looks good to me but it should be added in https://github.com/Telecominfraproject/wlan-ucentral-client instead.

ec-sam:
Yes, you are right. Let me do it again.

ec-sam:
Move to the correct repo (https://github.com/Telecominfraproject/wlan-ucentral-client)

PR #1151, opened 2026-06-30
rrmd: fix captive portal client removal and re-authentication flow

abhivyas1992:
rrmd: fix captive portal client removal and re-authentication flow
    Fixes: WIFI-15563
    Two issues in kick() prevented proper client removal and re-auth:
    1. !msg.ban_time treated ban_time=0 (immediate deauth, no roaming ban)
       as falsy, silently aborting the kick. Fix: guard on
       msg.ban_time == null instead, so 0 is accepted and passed through
       to hostapd as-is. hostapd already bans on every radio itself via
       hapd->iface->interfaces->for_each_interface() when global_ban is
       set, so no extra per-interface looping is needed on the rrmd side.
    2. spotfilter authorization was left intact after kick, so the client
       got internet access on reconnect without re-authenticating through
       the captive portal. Fix: call uspot.client_remove (to close RADIUS
       accounting/bookkeeping) and spotfilter.client_remove directly, so
       the BPF auth entry is cleared even if uspot's in-memory cache is
       stale.
    
  https://telecominfraproject.atlassian.net/browse/WIFI-15563

### DIFF
--- a/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
+++ b/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
@@ -221,7 +221,7 @@ return {
 	},
 
 	kick: function(msg) {
-		if (!msg.addr || !msg.ban_time || !msg.reason)
+		if (!msg.addr || msg.ban_time == null || !msg.reason)
 			return false;
 
 		if (!exists(stations, msg.addr))
--- a/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
+++ b/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
@@ -238,6 +238,35 @@ return {
 		/* tell hostapd to kick a station via ubus */
 		global.ubus.conn.call(`hostapd.${stations[msg.addr].device}`, 'del_client', payload);
 
+		/* clear captive portal auth state so client must re-authenticate after reconnect */
+		let uci = require('uci').cursor();
+		uci.foreach('uspot', 'uspot', function(s) {
+			if (s['.anonymous'])
+				return;
+			let ifnames = type(s.ifname) == 'array' ? s.ifname : [s.ifname];
+			for (let ifname in ifnames) {
+				if (ifname == stations[msg.addr].device) {
+					/*
+					 * go through uspot first so it can close out RADIUS
+					 * accounting (Acct-Stop) and its own bookkeeping, but
+					 * uspot only acts if the client is in *its* in-memory
+					 * cache. Also call spotfilter directly so the BPF
+					 * authorization entry is guaranteed to be cleared even
+					 * if that cache is stale (e.g. after a uspot restart).
+					 */
+					global.ubus.conn.call('uspot', 'client_remove', {
+						interface: s['.name'],
+						address: msg.addr
+					});
+					global.ubus.conn.call('spotfilter', 'client_remove', {
+						interface: s['.name'],
+						address: msg.addr
+					});
+					break;
+				}
+			}
+		});
+
 		return true;
 	},
 
--- a/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
+++ b/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
@@ -221,7 +221,7 @@ return {
 	},
 
 	kick: function(msg) {
-		if (!msg.addr || !msg.ban_time || !msg.reason)
+		if (!msg.addr || msg.ban_time == null || !msg.reason)
 			return false;
 
 		if (!exists(stations, msg.addr))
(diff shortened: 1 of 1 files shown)


tanyasingh-ec:
I think the base is wrong; please recheck

sujeet765:
base: 4.2.0 right ?

tanyasingh-ec:
There's a huge mismatch; please check. I don't think 256 commits need to be merged here

Please rebase based on staging-for-v4.2.0-LTS branch

feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
```diff
@@ -221,7 +221,7 @@ return {
 	},
 
 	kick: function(msg) {
-		if (!msg.addr || !msg.ban_time || !msg.reason)
+		if (!msg.addr || msg.ban_time == null || !msg.reason)
```
mkwacz, line 224: I don't understand? Why ban_time == 0 as a valid value here? The guard was changed from !msg.ban_time to msg.ban_time == null specifically to let 0 through, but line 234 (msg.ban_time || 30) immediately overwrites 0 with 30, so could you explain the intended use case for 0, since as written it never actually takes effect?

feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
```diff
@@ -231,12 +231,37 @@ return {
...
+		let device = stations[msg.addr].device;
+
+		/* kick from current device */
+		global.ubus.conn.call(`hostapd.${device}`, 'del_client', payload);
+
+		/* apply ban on all other interfaces to prevent roaming */
+		for (let d in global.local.status()) {
+			if (d != device)
+				global.ubus.conn.call(`hostapd.${d}`, 'del_client', payload);
+		}
```
venkatchimata: If you are banning on all the interfaces, you can just remove this 	global.ubus.conn.call(`hostapd.${device}`, 'del_client', payload);
 and  the d != device check can be avoided in the for loop.

venkatchimata: One thought : I think this is by design that the client will be kicked out on the specified interface. If you want to ban the client on other interfaces, then probably multiple calls to kick on the specified interface is needed. 

Banning on other interfaces too may break some existing stuff. Probably you can protect this loop with global_ban variable which was defined earlier.

venkatchimata (changes requested):
I am not sure if we need this 

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

sujeet765:
(1)   ban_time fixed-  ban_time: msg.ban_time now passes 0 through as-is.                                                                          
         Use case:                                                                                                                                   
-         0 = immediate deauth, no roaming ban;                                                                                                       
-         positive value = block reassociation for N seconds.
  
(2) only ban other radios when global_ban is set, to avoid breaking roam/band-steering

(3) uspot closes RADIUS accounting; spotfilter clear ensures BPF auth is reset even if uspot's cache is stale

venkatchimata:
@sujeet765  Please squash these three commits into one. They all seem to be part of the same issue. Please share your thoughts if you think the commits should be in the current state.

feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
venkatchimata: @sujeet765 A correction from my earlier comment. This is not needed. global is passed to the hostapd via del_client . hostapd handles global flag and bans the clients on all interfaces if it is set.

venkatchimata: So effectively no change is needed in terms of global ban .

venkatchimata (changes requested):

venkatchimata:
@sujeet765 In the commit, please add Fixes: WIFI-15563 above Signed-off line.
Please also fix the PR heading. You can make it "rrmd: fix captive portal client removal and re-authentication flow" and please update the PR description from the commit description.

feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
```diff
@@ -235,8 +235,39 @@ return {
 			global_ban: msg.global_ban || false,
 		};
 
+		let device = stations[msg.addr].device;
+
 		/* tell hostapd to kick a station via ubus */
-		global.ubus.conn.call(`hostapd.${stations[msg.addr].device}`, 'del_client', payload);
+		global.ubus.conn.call(`hostapd.${device}`, 'del_client', payload);
```
venkatchimata: Please revert these extra changes.. you dont need to change anything in the above two lines.

venkatchimata:
> @sujeet765 In the commit, please add Fixes: WIFI-15563 above Signed-off line. Please also fix the PR heading. You can make it "rrmd: fix captive portal client removal and re-authentication flow" and please update the PR description from the commit description.

This is not done yet.

feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
sujeet765: yes now above statements are reverted.

sujeet765:
@venkatchimata please review the final changes.

venkatchimata (approved):

sujeet765:
@tanyasingh-ec please review the code and merge.

tanyasingh-ec (changes requested):
Commit msg format is incorrect. Please follow guidelines.

feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
tanyasingh-ec: Has this been resolved? @mkwacz @abhivyas1992

mkwacz (approved):

tanyasingh-ec:
Will this be addressed for the `main` branch for 5.1 release?

sujeet765:
yes 
main for ->  5.1 release
and 4.2.0 LTS -> 4.2.x release

tanyasingh-ec:
@sujeet765 @abhivyas1992 I don't see any PR for the `main` branch
cc: @venkatchimata  @mkwacz

venkatchimata:
@sujeet765 Changes are ok. I reviewed this earlier. As Tanya said, please open the PR towards the main and once it is merged into the main, please cherry-pick the commits into staging-for-v4.2.0-LTS.

tanyasingh-ec (approved):

tanyasingh-ec:
@venkatchimata please approve this PR if all good

venkatchimata (approved):

PR #1152, opened 2026-06-30
ucentral-schema: fix HaLow center frequency shows wrong

ian77chen:
Root Cause:
The state reporting code calculates HaLow frequency incorrectly. It computes lower and upper edge frequencies using center_freq minus/plus bandwidth/2, then reports those two edge values as the frequency array. This does not match the expected format where channels and frequencies should be 1-to-1 mapped, same as 2.4G and 5G radios. The survey frequency variable also uses this incorrect center_freq value.

Solution:
Replace the edge frequency calculation with a per-channel frequency lookup from the S1G mapping table. Build both s1g_channels and s1g_frequencies arrays in parallel, so each channel maps to its correct S1G center frequency. Update the survey frequency to use the mapped S1G frequency directly. (This new version fix can fix all 1/2/4/8M HaLow bandwidth)

Fixes: WIFI-15501

### DIFF
--- a/feeds/ucentral/ucentral-schema/patches/004-fix-halow-freq-report.patch
+++ b/feeds/ucentral/ucentral-schema/patches/004-fix-halow-freq-report.patch
@@ -1,6 +1,6 @@
 --- a/system/state.uc
 +++ b/system/state.uc
-@@ -524,36 +524,31 @@
+@@ -524,36 +524,42 @@
  			let bandwidth = s1g_info.s1g_bw;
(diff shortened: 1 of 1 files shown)


SebastianHuang-ec (approved):
Looks ok to me

tanyasingh-ec (approved):

tanyasingh-ec:
Please resend PRs for https://github.com/Telecominfraproject/wlan-ucentral-schema/tree/main-v4.2.0-LTS directly

PR #1154, opened 2026-07-01
netifd, hostapd: bridge RADIUS dynamic VLAN clients on ath12k

mkwacz:
## What
Fix RADIUS-assigned dynamic VLAN (DVLAN) clients on ath12k getting no DHCP: the dynamically created AP_VLAN interface was never bridged.

## Why
- hostapd: the QCA nl80211 path creates and binds the AP_VLAN via `set_sta_vlan`, bypassing `vlan_if_add()` where the ubus `vlan_add` event is emitted, so userspace was never notified.
- netifd: `add_device` parses the AP_VLAN name (`phyX.Y-Z-vNNN`) as a VLAN chain, splitting at the first dot and failing to resolve it ("Not found").

## How
- hostapd: emit `vlan_add` from `ap_sta_bind_vlan()`.
- netifd: when an external device whose name contains a dot already exists as a netdev, use it directly instead of VLAN-parsing.

### DIFF
--- a/feeds/qca-wifi-7/hostapd/wlan-app-extns/hostapd/patches/zzz-t99-ubus-vlan-add-qca-dynamic-vlan.patch
+++ b/feeds/qca-wifi-7/hostapd/wlan-app-extns/hostapd/patches/zzz-t99-ubus-vlan-add-qca-dynamic-vlan.patch
@@ -1,0 +1,27 @@
+From: Marek Kwaczynski <marek@shasta.cloud>
+Subject: [PATCH] hostapd: emit ubus vlan_add for QCA driver-created dynamic AP_VLAN
+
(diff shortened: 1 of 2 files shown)


venkatchimata:
LGTM except the Signed-off-by:  missing inside the hostapd patch.

mkwacz:
> LGTM except the Signed-off-by: missing inside the hostapd patch.

Thanks, updated.

venkatchimata (approved):

tanyasingh-ec (changes requested):
Since both commits are for a single ticket, could you perhaps combine the two commits into one?

mkwacz:
> Since both commits are for a single ticket, could you perhaps combine the two commits into one?

It is not a good idea because one is related to netifd, second to hostapd.

tanyasingh-ec (approved):

PR #1160, opened 2026-07-03
ucentral-schema: Fix HaLow SSID frequency report

ian77chen:
Root Cause:
Under the morse driver, HaLow/S1G radios tune on borrowed 5GHz channels, so wif.frequency carries 5GHz values (e.g. 5560/5570). The SSID-level state reported these directly via ssid.frequency = uniq(wif.frequency). Since the cloud classifies band by frequency, it filed the HaLow SSID under 5GHz, so HaLow client/radio info did not show in Active Clients. (The radio-level state already remapped to S1G; the SSID path did not.)

Solution:
Added process_halow_ssid(), which applies the same map5GToS1G() conversion as the radio path for S1G/HaLow SSIDs: same channel-selection logic,
remapping the 5GHz frequency to real S1G frequency while keeping channel↔frequency 1:1 aligned. Non-HaLow SSIDs keep the original behaviour.

Fixes: WIFI-15593

### DIFF
--- a/feeds/ucentral/ucentral-schema/patches/007-fix-halow-ssid-freq-report.patch
+++ b/feeds/ucentral/ucentral-schema/patches/007-fix-halow-ssid-freq-report.patch
@@ -1,0 +1,54 @@
+--- a/system/state.uc
++++ b/system/state.uc
+@@ -820,7 +820,50 @@
+ 				ssid.ssid = wif.ssid || vap.config.mesh_id;
+ 				ssid.mode = wif.mode;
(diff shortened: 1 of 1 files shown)


tanyasingh-ec (approved):

tanyasingh-ec (approved):

tanyasingh-ec:
Please resend PRs for https://github.com/Telecominfraproject/wlan-ucentral-schema/tree/main-v4.2.0-LTS directly

PR #1164, opened 2026-07-03
qca-wifi-6/hostapd: fix incorrect psk_idx value

Andy79Yang:
hostapd_wpa_auth_get_psk()  walks the per-STA sta->psk list to find the PSK matching the 4-way-handshake MIC and records the matched position in sta->psk_idx, which mpskd maps back to the configured key/VLAN. The local psk_idx counter is initialised to 1 but never incremented while walking the list, so every match beyond the first entry is reported as index 1. A station whose key sits at list position >= 2 therefore receives the 2nd key's VLAN instead of its own.

Advance psk_idx as the list is walked so sta->psk_idx reflects the true position of the matching PSK.

This is the same fix of this PR
https://github.com/Telecominfraproject/wlan-ap/pull/1163

Fixes: WIFI-15354

### DIFF
--- a/feeds/qca-wifi-6/hostapd/patches/601-ucode_support.patch
+++ b/feeds/qca-wifi-6/hostapd/patches/601-ucode_support.patch
@@ -696,7 +696,7 @@
  #ifdef CONFIG_OWE
  	if ((hapd->conf->wpa_key_mgmt & WPA_KEY_MGMT_OWE) &&
  	    sta && sta->owe_pmk) {
-@@ -381,12 +382,17 @@ static const u8 * hostapd_wpa_auth_get_p
+@@ -381,13 +382,18 @@ static const u8 * hostapd_wpa_auth_get_p
  	 * returned psk which should not be returned again.
  	 * logic list (all hostapd_get_psk; all sta->psk)
  	 */
@@ -709,11 +709,13 @@
  		if (vlan_id)
  			*vlan_id = 0;
  		psk = sta->psk->psk;
+-		for (pos = sta->psk; pos; pos = pos->next) {
 +		if (vlan_id)
 +			sta->psk_idx = psk_idx;
- 		for (pos = sta->psk; pos; pos = pos->next) {
++		for (pos = sta->psk; pos; pos = pos->next, psk_idx++) {
  			if (pos->is_passphrase) {
  				pbkdf2_sha1(pos->passphrase,
+ 						hapd->conf->ssid.ssid,
 @@ -397,9 +403,13 @@ static const u8 * hostapd_wpa_auth_get_p
  			}
  			if (pos->psk == prev_psk) {


feeds/qca-wifi-6/hostapd/patches/zzz-sta-psk-idx-fix.patch
```diff
@@ -0,0 +1,10 @@
+--- a/src/ap/wpa_auth_glue.c
++++ b/src/ap/wpa_auth_glue.c
+@@ -413,6 +413,7 @@ static const u8 * hostapd_wpa_auth_get_psk(void *ctx, const u8 *addr,
+ 					sta->psk_idx = psk_idx + 1;
+ 				break;
+ 			}
++			psk_idx++;
```
venkatchimata: This looks good. Two cosmetic changes though.
1. Please increment psk_idx inside the for loop's update part as done here: https://github.com/Telecominfraproject/wlan-ap/blob/release/v4.2.0/feeds/qca-wifi-7/hostapd/patches/601-ucode_support.patch#L775 .  Other hostapd versions are doing this. It will be uniform across all the versions if we follow the same.
2. Instead of a separate patch for this change (as the change is one line addition), please modify 601-ucode_support.patch itself.

venkatchimata (changes requested):

feeds/qca-wifi-6/hostapd/patches/zzz-sta-psk-idx-fix.patch
Andy79Yang: Updated.

feeds/qca-wifi-6/hostapd/patches/601-ucode_support.patch
```diff
@@ -714,14 +714,15 @@
...
-@@ -397,9 +403,13 @@ static const u8 * hostapd_wpa_auth_get_p
+@@ -397,9 +403,14 @@ static const u8 * hostapd_wpa_auth_get_p
  			}
  			if (pos->psk == prev_psk) {
  				psk = pos->next ? pos->next->psk : NULL;
 +				if (vlan_id)
 +					sta->psk_idx = psk_idx + 1;
  				break;
  			}
++			psk_idx++;
```
venkatchimata: Can we move this increment inside the for loop? Patches for other platforms are doing that. I know the above works. Just want to make things uniform across the platforms.

Andy79Yang: Hi @venkatchimata 
It is already in the loop

Image

venkatchimata: @Andy79Yang Apologies if I did not communicate  correctly. Please refer this: https://github.com/Telecominfraproject/wlan-ap/blob/release/v4.2.0/feeds/qca-wifi-7/hostapd/patches/601-ucode_support.patch#L775 . psk_idx is updated inside the update part of the for loop.
image

Andy79Yang: Hi @venkatchimata 
Updated.

venkatchimata (approved):

feeds/qca-wifi-6/hostapd/patches/601-ucode_support.patch
venkatchimata: Thank you. Looks good to me.

SebastianHuang-ec (approved):
Looks good to me.

tanyasingh-ec:
DCO check failing. Commit needs to be revised

PR #1165, opened 2026-07-03
portal.uc: use null instead of NULL on handle_request error paths

venkatchimata:
Bare NULL at portal.uc:257 and :275 throws at runtime in uspot's
strict-mode uhttpd handler ("access to undeclared variable NULL"),
though it silently resolves in non-strict shells like plain `ucode`.
Lowercase null is the language literal and works in both; matches
line 115's `return null;` and line 117's `let reply = null;`.

Fixes: WIFI-15592
Signed-off-by: Venkat Chimata <venkat@nearhop.com>

### DIFF
--- a/feeds/ucentral/uspot/files/usr/share/uspot/portal.uc
+++ b/feeds/ucentral/uspot/files/usr/share/uspot/portal.uc
@@ -254,7 +254,7 @@ return {
 		if (!ctx.mac) {
 			this.syslog(ctx, 'failed to look up mac');
 			include('error.uc', ctx);
-			return NULL;
+			return null;
 		}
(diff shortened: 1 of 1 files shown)


tanyasingh-ec (approved):

mkwacz (approved):

tanyasingh-ec:
@venkatchimata  Can we perhaps shorten the commit msg to be more concise?

venkatchimata:
> @venkatchimata Can we perhaps shorten the commit msg to be more concise?

Agreed. I think the current one also should be fine. Probably from next commit, I will note this down. Please let me know.

venkatchimata:
> @venkatchimata Can we perhaps shorten the commit msg to be more concise?

Done.

PR #1166, opened 2026-07-03
mac80211: scan: fix ms/jiffies mismatch in passive scan overhead

mkwacz:
On mediatek/filogic (mac80211 backports 6.12.6) set the passive
channel dwell to scan_req->duration minus a fixed overhead
(IEEE80211_PASSIVE_MIN_CHANNEL_TIME, HZ/20 = 50ms at CONFIG_HZ=100)
to compensate for FW channel switch and wiphy_delayed_work scheduling.
Measured channel active time then tracks the requested duration to
within ~5ms for durations >= 100ms.

Fixes: WIFI-14822

### DIFF
--- a/feeds/mediatek/mac80211/patches/subsys/mtk-9905-mac80211-decrease-passive-scan-chan-time.patch
+++ b/feeds/mediatek/mac80211/patches/subsys/mtk-9905-mac80211-decrease-passive-scan-chan-time.patch
@@ -1,0 +1,28 @@
+Index: backports-6.12.6/net/mac80211/scan.c
+===================================================================
+--- backports-6.12.6.orig/net/mac80211/scan.c
++++ backports-6.12.6/net/mac80211/scan.c
+@@ -29,6 +29,9 @@
+ #define IEEE80211_CHANNEL_TIME (HZ / 33)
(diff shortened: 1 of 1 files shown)


venkatchimata (approved):

SebastianHuang-ec (commented):
This patch will not be applied to EAP111 because EAP111 use the "mediatek" feed instead of "mediatek-sdk" feed in the LTS branch. But this patch can still apply to other mediatek platforms, such as EAP112.

mkwacz:
It is only draft, I would like to validate sth on 4.2.0.

mkwacz:
> This patch will not be applied to EAP111 because EAP111 use the "mediatek" feed instead of "mediatek-sdk" feed in the LTS branch. But this patch can still apply to other mediatek platforms, such as EAP112.

Thanks for this clue.

mkwacz:
Added ucron fix to rebuild image for testing too.

SebastianHuang-ec (approved):

tanyasingh-ec (changes requested):
Local ucrun is already available in the 4.2.0-LTS branch. No need to include that commit

mkwacz:
> Local ucrun is already available in the 4.2.0-LTS branch. No need to include that commit

Agree, it was only due to preparing test image for validation. Patch rebased.

PR #1170, opened 2026-07-08
mediatek:add linux kernel 6.6 support for sonicfi_rap630w-211g

hehuimin-cybertan:
Update the Linux kernel for sonicfi_rap630w-211g from 5.4 to 6.6


venkatchimata (approved):

mkwacz:
Please share more context. Do you really need it 4.2? It looks like not hot fix but significant change.

mkwacz (changes requested):
Need explanation if we should really add it into 4.2 release.

hehuimin-cybertan:
We've updated the patch.
This PR is for supporting Sonicfi_RAP630W-211G(AX3000, MT7981+MT7976CN) aligned to upstream MediaTek Kernel6.6.  Currently, TIP4.2(kernel 6.6) has not included RAP630W-211G, we need to add it into 4.2.0-LTS.The similar device family (EAP111 and WPA588m) have already been added.
The patch specific to the 211g configuration and won't impact other projects. From our perspective，and as Arif mentioned as well, we’ve got the requirement that Sonicfi_RAP630W-211G on kernel 6.6 is necessary. Thanks!

mkwacz (approved):

mkwacz (changes requested):
Wait with merging, it must be common decision with https://github.com/Telecominfraproject/wlan-ap/pull/1171

SebastianHuang-ec (approved):

tanyasingh-ec:
> Wait with merging, it must be common decision with #1171

This PR is much simpler and is also similar to EAP111; low risk in merging @mkwacz

PR #1172, opened 2026-07-08
certificates & ucentral-schema: restore insta operational certs across factory reset (WIFI-15591)

abhivyas1992:
ucentral-schema: keep operational certs across factory reset and upgrade
    
    The uCentral app factory reset (cmd_factory.uc) and firmware upgrade
    (cmd_upgrade.uc) keep_redirector paths preserve the birth certs but not the
    operational cert/CA, so the AP loses its operational certificate. Add
    operational.pem and operational.ca to both keep lists via feed patches; the
    entries are archived only when present (existing fs.stat guard), so this is a
    no-op on devices without an operational cert.
    
    Complements the early_boot restore, which recovers the operational cert from
    the insta partition when only the birth key is kept.
    
    
    certificates: restore insta operational certs across factory reset
    
    On udaya,a5-id2 the EST-enrolled operational cert/CA are stored in the
    insta1/insta2 A/B partitions, but a factory reset left the AP without them.
    Three problems in the early-boot restore path, all fixed here:
    
    - mount_certs and store_certs shared a tar_part_lookup() that both read AND
      flipped the cert_part u-boot env, so the reader selected the opposite
      partition from where store_certs wrote. Make the reader map cert_part to
      the active partition without flipping.
    - mount_certs runs from early_boot at S09, before uboot-envtools generates
      /etc/fw_env.config (S95). A factory reset wipes that config, so fw_printenv
      returned empty and the wrong (empty) partition was picked. Generate the
      config first when it is missing.
    - early_boot skipped the restore whenever /etc/ucentral/key.pem existed. An
      app factory reset (keep_redirector) preserves the birth key.pem but not the
      operational cert, so mount_certs never ran. Return early only when fully
      provisioned (key.pem AND operational.pem present).
    
    Fixes: WIFI-15591
    Signed-off-by: abhishek vyas <abhishek.vyas@inventum.net>
    https://telecominfraproject.atlassian.net/browse/WIFI-15591


abhivyas1992:
@venkatchimata , we create new PR with updated PR Guidelines . now please check and provide your suggestion.

venkatchimata (approved):
LGTM.

abhivyas1992:
@ArifAlam , @tanyasingh-ec , @mkwacz Please review and merge PR.

mkwacz (changes requested):
@abhishekvyas1992 Please explain why this change is adding into release/v2.11.0 instead of main first. We need to keep align with main and cherry-pick changes to the releases.

- certificates fix (early_boot + mount_certs) is NOT in main -> land on main first, then cherry-pick.
- schema patches 002/003 are fine release-only: main already keeps operational certs via its newer schema pin, release/v2.11.0 pins the old 2022 schema that lacks it. Agreed for schema.

mkwacz (approved):

PR #1174, opened 2026-07-13
mediatek: mac80211: keep passive dwell floor when scan duration is unset

mkwacz:
When scan_req->duration is 0 (the default for all normal and roaming scans, nothing sets it), the two-way expression collapsed next_delay to 0, cutting passive channel dwell to roughly the FW/scheduling overhead (~40ms) and missing most beacons on passive and DFS channels. Restore the PASSIVE_CHANNEL_TIME floor for the duration==0 case while keeping the HZ/20 overhead compensation for explicit non-zero durations.

Fixes: WIFI-14822


mkwacz:
Sorry but I missed last version of patch which I validated. This change need to fix previous patch.

venkatchimata (approved):

SebastianHuang-ec (approved):

PR #1176, opened 2026-07-13
certificates: restore insta operational certs across factory reset

abhivyas1992:
certificates: restore insta operational certs across factory reset

Boards that store the EST-enrolled operational cert/CA in the insta1/insta2 A/B flash partitions could lose them after a factory reset. Three problems in the early-boot restore path, all fixed here:

- mount_certs and store_certs shared a tar_part_lookup() that both read AND flipped the cert_part u-boot env, so the reader selected the opposite partition from where store_certs wrote. Make mount_certs map cert_part to the active partition without flipping (store_certs keeps flipping).
- mount_certs runs from early_boot at S09, before uboot-envtools generates /etc/fw_env.config (S95). A factory reset wipes that config, so fw_printenv returned empty and the wrong (empty) partition was picked. Generate the config first when it is missing. Also fix the always-true [ -n "insta" ] guard to [ -n "$part" ].
- early_boot skipped the restore whenever /etc/ucentral/key.pem existed. An app factory reset (keep_redirector) preserves the birth key.pem but not the operational cert, so mount_certs never ran. Return early only when fully provisioned (key.pem AND operational.pem present).

Fixes: WIFI-15591
Signed-off-by: abhishek vyas <abhishek.vyas@inventum.net>

### DIFF
--- a/feeds/tip/certificates/files/etc/init.d/early_boot
+++ b/feeds/tip/certificates/files/etc/init.d/early_boot
@@ -30,7 +30,11 @@ boot() {
 		touch /tmp/squashfs
 	;;
 	esac
-	[ -f /etc/ucentral/key.pem ] && return
+	# Skip the restore only when fully provisioned. Checking key.pem alone was
+	# not enough: an app factory reset with keep_redirector (cmd_factory.uc)
+	# preserves key.pem but not operational.pem, which made this return early
+	# and never restore the operational cert stored in the insta partition.
+	[ -f /etc/ucentral/key.pem -a -f /etc/ucentral/operational.pem ] && return
 	/usr/bin/mount_certs
 	copy_certificates
 }
--- a/feeds/tip/certificates/files/usr/bin/mount_certs
+++ b/feeds/tip/certificates/files/usr/bin/mount_certs
@@ -8,16 +8,26 @@ check_certificates() {
 
 check_certificates 1
 
+# Read the currently-active cert partition. store_certs writes to the
+# alternate partition and then advances cert_part to point at it, so the
+# reader must return that same partition and must NOT flip cert_part
+# (flipping here made a factory-reset boot read the stale/empty slot).
+# fw_printenv needs /etc/fw_env.config, but the uboot-envtools uci-default
+# only generates it at S95done - long after this runs at S09early_boot. On
+# a factory-reset boot the config is gone, so cert_part reads empty and we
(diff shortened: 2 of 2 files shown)


abhivyas1992:
@mkwacz , as you suggested in PR https://github.com/Telecominfraproject/wlan-ap/pull/1176 , created new PR to add fix  in main branch.  Please Review. 

@venkatchimata, you please also review.

mkwacz (approved):

venkatchimata (approved):

sujeet765:
@tanyasingh-ec Please merge PR.

tanyasingh-ec (changes requested):
https://telecominfraproject.atlassian.net/browse/WIFI-15591 is targeted for 2.11, but this PR is for 5.1.0. Please check this.

Also, add the test steps and validation in the Jira ticket.

sujeet765:
> https://telecominfraproject.atlassian.net/browse/WIFI-15591 is targeted for 2.11, but this PR is for 5.1.0. Please check this.
> 
> Also, add the test steps and validation in the Jira ticket.

@tanyasingh-ec please read the last comment of PR #1172 (for 2.11.x)

sujeet765:
> https://telecominfraproject.atlassian.net/browse/WIFI-15591 is targeted for 2.11, but this PR is for 5.1.0. Please check this.
> 
> Also, add the test steps and validation in the Jira ticket.

@tanyasingh-ec, we originally created PR #1172, which included two fixes:

1. A certificate-related fix.
2. A uCentral schema patch for the 2.11 target.

During the review, @mkwacz suggested applying the certificate fix to the main branch first, before merging it into the 2.11.x target. Based on that feedback, we implemented the certificate fix separately in main and created this PR.

Do we need to create a new Jira ticket for this change, or can we continue tracking it under the existing ticket: WIFI-15591?

We have already added the 5.1.x target as an affected version in the same Jira.

venkatchimata:
I tested this on EAP105, mainly to check if factory resetting breaks anything. The AP cameup fine and got connected to the controller too.

mkwacz:
@tanyasingh-ec Are you planing to validate it in your setup before approving?

tanyasingh-ec:
> @tanyasingh-ec Are you planing to validate it in your setup before approving?

yes @mkwacz

tanyasingh-ec (approved):

PR #1178, opened 2026-07-14
rrmd: main fix captive portal client removal and re-authentication flow

abhivyas1992:
rrmd: main fix captive portal client removal and re-authentication flow
  
Two issues in kick() prevented proper client removal and re-auth:
 
1. !msg.ban_time treated ban_time=0 (immediate deauth, no roaming ban)
as falsy, silently aborting the kick. Fix: guard on
msg.ban_time == null instead, so 0 is accepted and passed through
to hostapd as-is. hostapd already bans on every radio itself via
hapd->iface->interfaces->for_each_interface() when global_ban is
set, so no extra per-interface looping is needed on the rrmd side.
 
2. spotfilter authorization was left intact after kick, so the client
got internet access on reconnect without re-authenticating through
the captive portal. Fix: call uspot.client_remove (to close RADIUS
accounting/bookkeeping) and spotfilter.client_remove directly, so
the BPF auth entry is cleared even if uspot's in-memory cache is
stale.
 
Fixes: WIFI-15563

Signed-off-by: Sujeet Kumar sujeet.kumar@inventum.net

### DIFF
--- a/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
+++ b/feeds/ucentral/rrmd/files/usr/share/rrmd/station.uc
@@ -221,7 +221,7 @@ return {
 	},
 
 	kick: function(msg) {
-		if (!msg.addr || !msg.ban_time || !msg.reason)
+		if (!msg.addr || msg.ban_time == null || !msg.reason)
 			return false;
 
 		if (!exists(stations, msg.addr))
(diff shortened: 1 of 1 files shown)


sujeet765:
@tanyasingh-ec @venkatchimata please review and merge PR

venkatchimata:
> @tanyasingh-ec @venkatchimata please review and merge PR

The code looks good.  I just updated the PR description.

venkatchimata (approved):

tanyasingh-ec (approved):

PR #1179, opened 2026-07-14
.github/scripts, .github/workflows: refine PR guideline checks and co…

tanyasingh-ec:
…mment output

The PR-guidelines workflow flagged some cases and missed a couple of rules from the contribution guides. Tighten it up:

- skip draft PRs so work-in-progress isn't nagged until ready
- report "check failed unexpectedly" when a check exits non-zero without emitting an annotation, instead of claiming no issues were found
- truncate overly long commit subjects in messages so a runaway subject line doesn't flood the annotations and PR comment
- make the Fixes: trailer check case-aware, nudging toward the canonical capital-F "Fixes: WIFI-#####" (still advisory)
- warn when a commit has no explanatory body (§2), ignoring blank lines and known trailers so mechanical commits and reverts don't false-warn
- route the PR-description ticket check through run_check so its error reaches stdout as an inline annotation, not only the sticky comment
- skip fork PRs, whose read-only token can't post the sticky comment

### DIFF
--- a/.github/scripts/check-commits.sh
+++ b/.github/scripts/check-commits.sh
@@ -12,6 +12,7 @@
 #   - missing Fixes: WIFI-#### trailer        (§2: "where applicable")
 #   - branch name not WIFI-<num>-<desc>       (A1; fork branches vary)
 #   - subject line longer than 72 chars
+#   - no explanatory body                     (§2; mechanical commits may skip)
 
 set -euo pipefail
 
@@ -24,6 +25,17 @@ range="${BASE_SHA}..${HEAD_SHA}"
 
 err()  { echo "::error::$*";   fail=1; }
 warn() { echo "::warning::$*"; }
+
+# Shorten a subject for display in messages so a runaway subject line doesn't
+# flood the annotations / PR comment. Reports the full length separately.
+trunc() {
+  local s="$1" max=80
+  if [ "${#s}" -gt "$max" ]; then
+    printf '%s…' "${s:0:max}"
+  else
+    printf '%s' "$s"
+  fi
+}
 
 # ---------------------------------------------------------------
 # 1. No merge commits — history must replay cleanly onto main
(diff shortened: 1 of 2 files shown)


mkwacz (approved):

mkwacz:
Small issue in the PR-description check (section 3): the whole block is redirected to `sections.md`, so the `::error::` line goes into the sticky comment as raw text and never reaches stdout — so GitHub shows **no inline annotation** for it. The other checks avoid this because `run_check` prints `::error::` to stdout and puts only the `- ❌` line in the comment.

Simplest fix is to reuse `run_check`:

```sh
check_pr_desc() {
  echo "$PR_BODY" | grep -qE '\b(WIFI|WLAN)-[0-9]+\b' && return 0
  echo "::error::PR description must reference the ticket (WIFI-##### / WLAN-####)."
  return 1
}
run_check "PR description (ticket link)" check_pr_desc

If you prefer to keep it inline, just move the echo "::error::..." out of the >> sections.md block so it goes to stdout, and keep only the - ❌ ... line in the comment.

venkatchimata (approved):

tanyasingh-ec:
> Small issue in the PR-description check (section 3): the whole block is redirected to `sections.md`, so the `::error::` line goes into the sticky comment as raw text and never reaches stdout — so GitHub shows **no inline annotation** for it. The other checks avoid this because `run_check` prints `::error::` to stdout and puts only the `- ❌` line in the comment.
> 
> Simplest fix is to reuse `run_check`:
> 
> ```shell
> check_pr_desc() {
>   echo "$PR_BODY" | grep -qE '\b(WIFI|WLAN)-[0-9]+\b' && return 0
>   echo "::error::PR description must reference the ticket (WIFI-##### / WLAN-####)."
>   return 1
> }
> run_check "PR description (ticket link)" check_pr_desc
> 
> If you prefer to keep it inline, just move the echo "::error::..." out of the >> sections.md block so it goes to stdout, and keep only the - ❌ ... line in the comment.
> ```

Thanks for the suggestion. Changes made. @mkwacz 
@jaspreetsachdev  Please take a look as well

mkwacz (approved):

PR #1180, opened 2026-07-15
ucentral-client: update to latest HEAD (cb48fe2)

tanyasingh-ec:
cb48fe2 tls: validate hostname against subjectAltName


mkwacz:
Not a blocker, this follows the existing `…: update to latest HEAD` convention. I'd just prefer the short commit hash in the subject, e.g.:

`ucentral-client: update to latest HEAD (cb48fe2)`

"latest HEAD" alone is time-relative and doesn't record what was pinned. The hash also saves work later when we step up to the latest TIP sources: from `git log` you can see at a glance which revision each bump moved to, instead of digging it out of the Makefile commit by commit.

tanyasingh-ec:
> Not a blocker, this follows the existing `…: update to latest HEAD` convention. I'd just prefer the short commit hash in the subject, e.g.:
> 
> `ucentral-client: update to latest HEAD (cb48fe2)`
> 
> "latest HEAD" alone is time-relative and doesn't record what was pinned. The hash also saves work later when we step up to the latest TIP sources: from `git log` you can see at a glance which revision each bump moved to, instead of digging it out of the Makefile commit by commit.

For this ticket specifically, only one commit hash is added, but in some cases multiple commit hashes need to be added, so it may not be so feasible to add them all to the commit title (example: https://github.com/Telecominfraproject/wlan-ap/commit/3bd798e3770395eecacb114ba3113f9c7b68c37c)

Let me know your thoughts @mkwacz

mkwacz:
> > Not a blocker, this follows the existing `…: update to latest HEAD` convention. I'd just prefer the short commit hash in the subject, e.g.:
> > `ucentral-client: update to latest HEAD (cb48fe2)`
> > "latest HEAD" alone is time-relative and doesn't record what was pinned. The hash also saves work later when we step up to the latest TIP sources: from `git log` you can see at a glance which revision each bump moved to, instead of digging it out of the Makefile commit by commit.
> 
> For this ticket specifically, only one commit hash is added, but in some cases multiple commit hashes need to be added, so it may not be so feasible to add them all to the commit title (example: [3bd798e](https://github.com/Telecominfraproject/wlan-ap/commit/3bd798e3770395eecacb114ba3113f9c7b68c37c))
> 
> Let me know your thoughts @mkwacz

I would like to add hash to latest HEAD to see which real hash is used. 
Based on example what you shared only top hash should be used: 23acec3
ucentral-schema: update to latest HEAD (23acec3)
23acec3 state: fix HaLow center frequency in state message
df8443b captive: add walled-garden firewall allow rules before drop rule
b6a8b79 state: fix HaLow interface frequency report in the state message

carsten989 (approved):

tanyasingh-ec:
> > > Not a blocker, this follows the existing `…: update to latest HEAD` convention. I'd just prefer the short commit hash in the subject, e.g.:
> > > `ucentral-client: update to latest HEAD (cb48fe2)`
> > > "latest HEAD" alone is time-relative and doesn't record what was pinned. The hash also saves work later when we step up to the latest TIP sources: from `git log` you can see at a glance which revision each bump moved to, instead of digging it out of the Makefile commit by commit.
> > 
> > 
> > For this ticket specifically, only one commit hash is added, but in some cases multiple commit hashes need to be added, so it may not be so feasible to add them all to the commit title (example: [3bd798e](https://github.com/Telecominfraproject/wlan-ap/commit/3bd798e3770395eecacb114ba3113f9c7b68c37c))
> > Let me know your thoughts @mkwacz
> 
> I would like to add hash to latest HEAD to see which real hash is used. Based on example what you shared only top hash should be used: 23acec3 ucentral-schema: update to latest HEAD (23acec3) 23acec3 state: fix HaLow center frequency in state message df8443b captive: add walled-garden firewall allow rules before drop rule b6a8b79 state: fix HaLow interface frequency report in the state message

Noted and updated @mkwacz

venkatchimata (approved):

mkwacz (approved):

tanyasingh-ec:
the build is failing due to incorrect mirror hash. I will update that and then merge the PR

PR #1181, opened 2026-07-15
hostapd: fix MUEDCA user-config value in mac80211.sh only applied to 1 Radio

Niko525620:
Problem: When users manually modify the following four attributes in /lib/netifd/wireless/mac80211.sh:
- he_mu_edca_ac_be_timer
- he_mu_edca_ac_bk_timer
- he_mu_edca_ac_vi_timer
- he_mu_edca_ac_vo_timer

from the default value 255 to X, the actual MU EDCA Timer field in 802.11 beacon/probe response frames remains 0xff (255) instead of the user‑set value X.

Root cause:
This happens because hostapd unconditionally overwrites these timers with the default value (255) during beacon construction, ignoring any user modifications.

Solution:
Fix this by adding a check in hostapd: if the user‑supplied value is not equal to the default (255), skip the default overwrite and preserve the user configuration.

Tests: Successfully verified on RAP710C-341X

Fixes: WIFI-15466


mkwacz (approved):
Change looks good for me.

venkatchimata (approved):

PR #1171, opened 2026-07-08
mediatek:add linux kernel 6.6 support for sonicfi_rap63xc-211g 

hehuimin-cybertan:
Update the Linux kernel for sonicfi_rap63xc-211g from 5.4 to 6.6 
Porting mediatek/generic/image/ and mediatek/ramips/ from OpenWRT 24.10

PR #1183, opened 2026-07-16
mac80211:circular dependency between feeds/mediatek/mac80211 and ucode

hehuimin-cybertan:
The root cause is a circular dependency between feeds/mediatek/mac80211 and ucode: mac80211->hostapd->ucode, while ucode also depends on mac80211. As a result, the build system drops this dependency and chooses to compile ucode first, which leads to a missing nl80211.h.
To fix it, remove the hostapd dependency part from the Makefile in feeds/mediatek/mac80211

Fixes: WIFI-15599


tanyasingh-ec:
1. I will wait for the EAP111 firmware to build successfully and check the hostapd package
2. If all is good, then I will merge this PR and https://github.com/Telecominfraproject/wlan-ap/pull/1171 

cc: @venkatchimata @mkwacz @ArifAlam

PR #1184, opened 2026-07-16
ath11k: add patch to fix inter-radio roaming on IPQ807x_v5.4

joseGarciaDoblado:
Add a-106 patch that fixes some bugs causing random WiFi client disconnections on dual-radio APs after roaming between bands:

1. ar mismatch: peer_delete was called with the new radio's ar but the old radio's vdev_id, causing the completion signal to arrive on the wrong radio and timeout every time. Fixed by resolving the correct radio via ath11k_mac_get_ar_by_vdev_id before deletion.

2. Double teardown race: peer_create and mac_op_sta_state NOTEXIST could both delete the same peer concurrently, causing num_peers underflow and rhashtable corruption. Fixed by using an atomic claim mechanism with the existing delete_in_progress flag.

3. Peer-create must wait for a concurrent teardown: mac80211 does not retry a failed sta insertion: an error from ath11k_peer_create() propagates through sta_state(NOTEXIST->NONE) to NL80211_CMD_NEW_STATION, and hostapd rejects the (re)association with status 17.

Fixes: WIFI-15520

### DIFF
--- a/feeds/ipq807x_v5.4/mac80211/patches/pending/a-106-fix-ar-mismatch-cross-radio-peer-delete.patch
+++ b/feeds/ipq807x_v5.4/mac80211/patches/pending/a-106-fix-ar-mismatch-cross-radio-peer-delete.patch
@@ -1,0 +1,396 @@
+From: Jose Garcia Doblado <jose.garciadoblado@joindigital.com>
+Subject: [PATCH] wifi: ath11k: fix ar mismatch and double-teardown race in
+ cross-radio peer delete
+
+When a station roams between radios on the same AP, ath11k_peer_create()
+finds the existing peer on the old vdev and removes it before creating
+the new one:
+
+    ath11k_dp_peer_cleanup(ar, vdev_id, param->peer_addr);
+    ath11k_peer_delete(ar, vdev_id, param->peer_addr);
+
+`ar` here is the new radio performing the create, while `vdev_id` belongs
+to the old vdev on a different radio. This has two bugs, and the naive
+fix (fail peer_create when another path is already tearing the peer
+down) introduces a third problem, observed in the field as FT roaming
+silently degrading to full re-authentication. All three are addressed
+here.
+
+1) ar mismatch in ath11k_dp_peer_cleanup() and ath11k_peer_delete()
+
+ath11k_wait_for_peer_delete_done() waits on ar->peer_delete_done.
+ath11k_peer_delete_resp_event() (wmi.c) resolves the owning radio via
+ath11k_mac_get_ar_by_vdev_id(ab, vdev_id) and signals that radio's
+peer_delete_done. For a cross-radio roam these are two different struct
+ath11k instances, so the completion being waited on is never the one the
+firmware response signals - the wait always times out. Similarly,
+ath11k_dp_peer_cleanup() uses ar-local data-path state; calling it with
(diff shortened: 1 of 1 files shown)


mkwacz:
Thanks for this. The analysis is good and the cross-radio `ar` mismatch is a real bug we hit too.

My worry is timing, not correctness. We spent the whole last release making this peer create/delete path stable, and it is still fragile. This is a big change to add to it late in the current release without a long soak. One concrete point: the 1500 ms claim timeout can run out before a slow WMI `peer_delete` finishes, so the waiting `peer_create` gives up with `-EBUSY` and brings back the same association failure this is trying to fix.

I would prefer to keep it out of this release and give it a proper soak in the next one.

Btw: build is failed!

mkwacz (changes requested):
Build is failed. Change too risk.

joseGarciaDoblado:
@mkwacz @tanyasingh-ec @venkatchimata

The build is failing for **MediaTek** devices. I recommend reviewing the pipeline or the code for these device, since the patch only affects the ipq807x_v5.4 feed (the patch is located in **feeds/ipq807x_v5.4/mac80211/patches/pending**) and therefore has nothing to do with MediaTek devices. It doesn't even compile.

The **1500 ms** is a safety ceiling against a hung firmware, not the actual wait time on the normal path. The mechanism uses a Linux completion, which is fully event-driven: the moment the firmware sends the WMI peer_delete response, the completion is signalled and the waiting peer_create unblocks immediately. There is no polling and no sleeping for 1500 ms.

In practice a WMI peer_delete response comes back in tens of milliseconds, well under 100 ms even under load. The incoming peer_create waits for that real completion time, not for the timeout. The 1500 ms is only reached if the firmware stops responding entirely, which is already a hard failure where the whole stack is in trouble regardless of this change.

The alternative — the current bug — is actually worse in that same scenario: without the claim, the peer_create fires the WMI before the previous peer_delete has finished, the firmware receives two conflicting operations on the same peer_id, and the association fails anyway, but now with corrupted state instead of a clean -EBUSY.

So the change adds no latency to the normal roaming path. It only serialises what should already be serialised. The timeout being 1500 ms is deliberately conservative precisely so it never triggers in normal operation

@tanyasingh-ec has tested the patch in LandForge and can share the test report if necessary.

venkatchimata (approved):

mkwacz:
@ArifAlam Please review, since you also spent a lot of time stabilizing the roaming path in the 4.2 release. @tanyasingh-ec Please share the roaming results, especially for roaming between bands on the same AP
Currently unblocking merging because build passed.

mkwacz (commented):
Build passed, unblock merging.

tanyasingh-ec:
> @ArifAlam Please review, since you also spent a lot of time stabilizing the roaming path in the 4.2 release. @tanyasingh-ec Please share the roaming results, especially for roaming between bands on the same AP Currently unblocking merging because build passed.

@mkwacz @venkatchimata @ArifAlam Test results added https://telecominfraproject.atlassian.net/browse/WIFI-15520?focusedCommentId=63440

Please take a look

tanyasingh-ec (approved):

tanyasingh-ec:
Same as https://github.com/Telecominfraproject/wlan-ap/pull/1214

PR #1191, opened 2026-07-20
certificates: fix certificate partition unmounted on first install

Javen829:
Problem:
certificate partition is umounted when installing the certificates for the first time on EAP104 and EAP105.

Root Cause:
In the initial state, the primary partition has already created a volume named certificates, while the backup partition has not yet created such a volume (since the backup partition has never been booted, the /lib/preinit/75_certificates script has not been executed to create the certificates volume). If an attempt is made to mount the certificates volume from the backup partition, the mount operation will fail.Because the certificate files cannot be found, the script executes umount /certificates. In effect, this unmounts the /certificates mount point from the primary partition.

Solution:
Only execute check_certificates and umount /certificates after the mount of the backup partition's certificates volume has succeeded. This prevents the script from executing umount /certificates when the mount from the backup partition fails, which would otherwise unmount the /certificates mount point from the primary partition.

Tests: Successfully verified on RAP750W-311A

Fixes: WIFI-15363

### DIFF
--- a/feeds/tip/certificates/files/usr/bin/mount_certs
+++ b/feeds/tip/certificates/files/usr/bin/mount_certs
@@ -160,9 +160,11 @@ MTD=$(find_mtd_index $PART_NAME)
 [ -z "$MTD" ] && return 1
 
 ubiattach -m $MTD -d 3
-[ -e /dev/ubi3 ] && mount -t ubifs ubi3:certificates /certificates
-
-check_certificates || {
-	umount /certificates
-	check_certificates
-}
+if [ -e /dev/ubi3 ] ; then
+	if mount -t ubifs ubi3:certificates /certificates; then
+		check_certificates || {
+			umount /certificates
+			check_certificates
+		}
+	fi
+fi


venkatchimata:
@Javen829 Please see my comment on the JIRA ticket.

feeds/tip/certificates/files/usr/bin/mount_certs
```diff
@@ -160,9 +160,12 @@ MTD=$(find_mtd_index $PART_NAME)
...
-	check_certificates
-}
+if [ -e /dev/ubi3 ] ; then
+	if mount -t ubifs ubi3:certificates /certificates; then
+		check_certificates || {
+			umount /certificates
+			check_certificates
+		}
+	fi
+	ubidetach -d 3 2>/dev/null
```
venkatchimata: This change looks ok. however there is no mention of ubidetach. Could you please add few lines about it too.

venkatchimata:
Could you please test on few more boards? Testing on just one board might not be sufficient.

tanyasingh-ec:
> Could you please test on few more boards? Testing on just one board might not be sufficient.

Testing on Edgecore APs, shall share results today

feeds/tip/certificates/files/usr/bin/mount_certs
Javen829: We consider that it should run ubidetach if the mount fails or check_certificates fails. And check_certificates success will not run ubidetach.  Of course, it does not matter if there is no ubidetach.

venkatchimata: If it does n't matter, please remove ubidetach. If it is needed, please add and update the commit description.

Javen829: OK, Thank you!

tanyasingh-ec (changes requested):
remove `ubidetach`

venkatchimata (approved):
@tanyasingh-ec Given the sensitivity of the change, could you please verify things again (post the removal of ubidetach).

tanyasingh-ec:
> @tanyasingh-ec Given the sensitivity of the change, could you please verify things again (post the removal of ubidetach).

Will test. Merging shall be blocked until retest 
@Javen829  Please also test on your side

tanyasingh-ec (approved):
Works for Edgecore APs
Ref: https://telecominfraproject.atlassian.net/browse/WIFI-15363?focusedCommentId=62788

mkwacz (approved):

PR #1153, opened 2026-07-01
qca-wifi-7: change Sonicwifi RAP750W-311A/RAP750E-S/RAP750E-H to use UBIFS and add anti-rollback

crychic-cybertan:
We need to merge PR[#1027](https://github.com/Telecominfraproject/wlan-ap/pull/1027) and PR[#1035](https://github.com/Telecominfraproject/wlan-ap/pull/1035) into the v4.2.x branch.

The added changes are listed below:
1. Fix the certificate mount failure issue on RAP750W-311A/RAP750E-S/RAP750E-H devices within v4.2.x releases caused by incorrect certificate partition formatting.

2. The certificate partition format has been updated for RAP750W-311A/RAP750E-S/RAP750E-H and RAP7110c-341X. Downgrading firmware to older versions will lead to certificate partition mount failures, so we need to implement version restrictions for firmware upgrades.

Test Results:
1. After applying the new changes, RAP750W-311A/RAP750E-S/RAP750E-H can mount the certificate partition normally.

2. Upgrading to firmware versions lower than the restricted version, or firmware without version information but built earlier than the cutoff build date, will fail. Upgrades to firmware versions higher than the restricted version will succeed. Firmware without version information but with a build date later than the cutoff build date can also be upgraded successfully.

PR #1195, opened 2026-07-22
qca-wifi-7: update anti-rollback limit ver for 750W-311A/750E-S/750E-H

crychic-cybertan:
set anti-rollback limit to LIMIT_VER=4.2.6 / LIMIT_BUILD_DATE=20260717 (the first UBIFS release for these boards, per https://github.com/Telecominfraproject/wlan-ap/pull/1153)

Test Results:

Upgrading to firmware versions lower than the restricted version, or firmware without version information but built earlier than the cutoff build date, will fail. Upgrades to firmware versions higher than the restricted version will succeed. Firmware without version information but with a build date later than the cutoff build date can also be upgraded successfully.

Fixes: WIFI-15401

### DIFF
--- a/feeds/qca-wifi-7/ipq53xx/base-files/lib/upgrade/platform.sh
+++ b/feeds/qca-wifi-7/ipq53xx/base-files/lib/upgrade/platform.sh
@@ -225,8 +225,8 @@ platform_check_image() {
 		local info_output=$(get_firmware_info "$1")
 		FW_VER="${info_output%%|*}"
 		FW_BUILD_DATE="${info_output#*|}"
-		local LIMIT_VER="5.0.0"
-		local LIMIT_BUILD_DATE="20260603"
+		local LIMIT_VER="4.2.6"
+		local LIMIT_BUILD_DATE="20260717"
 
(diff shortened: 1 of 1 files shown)


mkwacz:
Change looks fine to me. Could you add the actual values to the commit message so the limit is visible without opening the diff, for example: "set anti-rollback limit to LIMIT_VER=4.2.6 / LIMIT_BUILD_DATE=20260717 (the first UBIFS release for these boards, per #1153)". Right now the message says the limit is updated but not to what, so it is hard to review at a glance.

crychic-cybertan:
> Change looks fine to me. Could you add the actual values to the commit message so the limit is visible without opening the diff, for example: "set anti-rollback limit to LIMIT_VER=4.2.6 / LIMIT_BUILD_DATE=20260717 (the first UBIFS release for these boards, per #1153)". Right now the message says the limit is updated but not to what, so it is hard to review at a glance.

@mkwacz Commit message and PR details updated, please re-review.

venkatchimata (approved):

mkwacz (approved):

PR #1196, opened 2026-07-22
mediatek: wifi-scripts: add psk2-radius encryption

SebastianHuang-ec:
Add the support for psk2-radius encryption.
Also set "radius_require_message_authenticator" to 0 by default.

Fixes: WIFI-15603

### DIFF
--- a/feeds/mediatek/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json
+++ b/feeds/mediatek/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json
@@ -931,6 +931,12 @@
 			"description": "Request Chargeable-User-Identity (RFC 4372)",
 			"type": "boolean"
 		},
+		"radius_require_message_authenticator": {
+			"description": "Require the Message-Authenticator attribute in received RADIUS packets (CVE-2024-3596 mitigation). Disabled by default to interoperate with RADIUS servers that omit it on non-EAP (MAC/PSK) responses",
+			"type": "number",
+			"enum": [ 0, 1 ],
+			"default": 0
+		},
 		"reassociation_deadline": {
 			"description": "Reassociation deadline in time units",
 			"type": "number",
--- a/feeds/mediatek/wifi-scripts/files-ucode/usr/share/ucode/wifi/ap.uc
+++ b/feeds/mediatek/wifi-scripts/files-ucode/usr/share/ucode/wifi/ap.uc
@@ -63,6 +63,7 @@ function iface_setup(config) {
 		'ocv', 'multicast_to_unicast', 'preamble', 'proxy_arp', 'per_sta_vif', 'mbo',
 		'bss_transition', 'wnm_sleep_mode', 'wnm_sleep_mode_no_keys', 'qos_map_set', 'max_listen_int',
 		'dtim_period', 'wmm_enabled', 'start_disabled', 'na_mcast_to_ucast',
+		'radius_require_message_authenticator',
 	]);
 }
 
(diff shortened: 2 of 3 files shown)


tanyasingh-ec (approved):

venkatchimata (approved):

mkwacz (approved):

feeds/mediatek/wifi-scripts/files-ucode/usr/share/ucode/wifi/ap.uc
```diff
@@ -185,6 +186,22 @@ function iface_auth_type(config) {
...
+		/* WPA2-PSK, but the PSK is fetched from the RADIUS server during
+		 * the 4-way handshake so clients can be centrally tracked. */
+		set_default(config, 'ieee80211w', 1);
+		config.vlan_possible = 1;
+		config.wpa_psk_radius = 3;
+		config.macaddr_acl = 2;
+
+		iface_authentication_server(config);
+
+		if (config.radius_das_client && config.radius_das_secret) {
```
mkwacz, line 201: export function set_default(dict, key, value) {
      if (dict[key] == null)
              dict[key] = value;
};

radius_das_client reassignment is a no-op.
if (config.radius_das_client && config.radius_das_secret) {
     set_default(config, 'radius_das_port', 3799);
     set_default(config, 'radius_das_client', `${config.radius_das_client} ${config.radius_das_secret}`);
 }
 set_default only assigns when the key is null, but radius_das_client is guaranteed non-null inside this if, so the "client secret" concatenation never runs and hostapd gets radius_das_client without its shared secret.

mkwacz (changes requested):

feeds/mediatek/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json
```diff
@@ -931,6 +931,12 @@
 			"description": "Request Chargeable-User-Identity (RFC 4372)",
 			"type": "boolean"
 		},
+		"radius_require_message_authenticator": {
```
mkwacz, line 934: Please confirm that default value is really working.
append_vars only emits keys that are present in the config, and validate.uc only validates values, it does not inject schema defaults into the config. So when a user sets encryption 'psk2-radius' without explicitly setting radius_require_message_authenticator, no radius_require_message_authenticator=0 line is written, and hostapd falls back to its own post-CVE-2024-3596 default of 1 (require). RADIUS servers that omit Message-Authenticator on PSK/MAC responses (for example PacketFence) then have their Access-Accept/Reject silently dropped, which is the exact interop problem this is meant to prevent.

Please set it explicitly in the psk2-radius branch of ap.uc, next to the existing set_default(config, 'ieee80211w', 1): set_default(config, 'radius_require_message_authenticator', 0);

mkwacz (changes requested):

feeds/mediatek/wifi-scripts/files-ucode/usr/share/ucode/wifi/ap.uc
SebastianHuang-ec: Thanks for pointing this out. Will modify the implementation for this.

feeds/mediatek/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json
SebastianHuang-ec: I've tested and the default value works for every BSS. However, I agree that it will be clearer if it is set explicitly. I will add it.

SebastianHuang-ec:
@mkwacz PR updated as suggested.

PR #1199, opened 2026-07-23
mediatek: mt76: enable wed_enable=1 for mt7915e on sonicfi_rap630w_211g

lyc-cybertan:
After adapting to the MTDK platform WiFi6 firmware, hardware acceleration failed to offload traffic because the WED function was not enabled. During 5G RvR testing, the LAN to WiFi forwarding rate was too low to meet specs.

In the upstream MT76 driver, WED is an optional hardware acceleration component and is disabled by default. The wed_enable module parameter must be manually configured to initialize the hardware offloading path, reducing CPU soft forwarding performance loss.

Modify package/kernel/mt76/Makefile to inject the mt7915e module parameter wed_enable=1 specifically for the VNet sonicfi_rap630w_211g target model. This enables the hardware traffic offloading path during firmware compilation, allowing LAN to WiFi traffic to be handled by WED.

Fixes: WIFI-15621

Test Results:
After compiling and deploying the firmware, retesting showed a significant improvement in LAN → WiFi direction transmission rate in the 5G RvR scenario, and all performance indicators met the testing requirements


venkatchimata (approved):

tanyasingh-ec:
Not a blocker but reminder to rebase your staging branches so that the guidelines are up-to-date

tanyasingh-ec (approved):

PR #1215, opened 2026-07-29
Previously GCMP-256 was enabled for both HE and EHT on 6GHz band.

Cyrusx7:
This change limits it to EHT only, as HE does not require GCMP-256.

Tested on WF196 with 6GHz radio, verified GCMP-256 is only applied under EHT mode and not under HE mode.

Fixes: WIFI-15627

### DIFF
--- a/patches/0104-netifd-limit-gcmp-256-to-EHT-only.patch
+++ b/patches/0104-netifd-limit-gcmp-256-to-EHT-only.patch
@@ -1,0 +1,37 @@
+From 386e28bccf11978022a461abdc438af2e4eb691b Mon Sep 17 00:00:00 2001
+From: ruanyaoyu <ruanyaoyu@actiontec.com>
+Date: Wed, 29 Jul 2026 18:28:43 +0800
+Subject: [PATCH] netifd: limit GCMP-256 cipher to EHT only on 6GHz
+
+Previously GCMP-256 was enabled for both HE and EHT on 6GHz band.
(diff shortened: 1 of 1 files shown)


mkwacz (approved):

tanyasingh-ec:
Waiting for test validation before merge

mkwacz (approved):

mkwacz (commented):
The change itself and its location look right to me. 

Note:
The header fields are placeholders. From 0000000000000000000000000000000000000000 and index 0000000000..1111111111 are not real object names. 
Please regenerate the file with git format-patch, or with quilt using QUILT_NO_DIFF_INDEX=1 so no index line is emitted at all. A real index or no index are both fine, a made up one is the only problematic case.

ArifAlam:
Verified that this fix works.

tanyasingh-ec:
> The change itself and its location look right to me.
> 
> Note: The header fields are placeholders. From 0000000000000000000000000000000000000000 and index 0000000000..1111111111 are not real object names. Please regenerate the file with git format-patch, or with quilt using QUILT_NO_DIFF_INDEX=1 so no index line is emitted at all. A real index or no index are both fine, a made up one is the only problematic case.

@Cyrusx7 Please fix this

Cyrusx7:
I have fixed these issues. Please help review and merge

mkwacz (approved):

SebastianHuang-ec (approved):

PR #1216, opened 2026-07-29
mac80211: enable built-in MLME/STA/HT/PS/counters debug logging

abhivyas1992:
Problem:
    IPQ-based boards on the ipq807x_v5.4 feed (e.g. udaya-a6-id2) have
    significantly less kernel-log visibility into client connect/
    disconnect/aggregation behavior than boards on other feeds, since
    mac80211's built-in diagnostic logging is compiled out by default.
    Troubleshooting connectivity issues requires reasoning about mac80211
    internals blind, with no narration of what it's actually doing.
    Root Cause:
    MAC80211_DEBUG_MENU, MAC80211_MLME_DEBUG, MAC80211_STA_DEBUG,
    MAC80211_HT_DEBUG, MAC80211_PS_DEBUG, and MAC80211_DEBUG_COUNTERS are
    not enabled in this package's config-y list, unlike other mac80211
    packages in this repo (e.g. feeds/mediatek-sdk) which already ship
    with this logging on.
    Fix:
    Add all six symbols to config-y. None of these add new WiFi
    functionality -- authentication, aggregation, and power-save already
    work identically whether these are enabled or not. The four _DEBUG
    options turn on log messages narrating what mac80211 already does
    internally at each connection stage; MAC80211_DEBUG_COUNTERS exposes
    passive statistics counters via debugfs for on-demand inspection.
    Tests: 
	Tested on udaya-a6-id2, verified full connect/disconnect
    lifecycle logging (STA allocation, state transitions, aggregation
    negotiation and teardown) now appears in logread, matching boards
    that already had this logging enabled.
    Fixes: WIFI-15630


venkatchimata:
Changes look ok. Do we need this for v4.2?

venkatchimata:
> Changes look ok. Do we need this for v4.2?

@tanyasingh-ec ^^

tanyasingh-ec:
@abhivyas1992 I recommend enabling these only for Udaya APs in the profile, as this is an Inventum-specific requirement for now
cc: @venkatchimata

mkwacz (changes requested):
I don't see reason to add it into 4.2 release. My recommendation is rejecting this PR.

tanyasingh-ec:
> I don't see reason to add it into 4.2 release. My recommendation is rejecting this PR.

I agree; this is a 'feature,' not an improvement or fix.

PR #1219, opened 2026-07-30
haLow: Add HaLow (802.11ah / S1G) support for USB HaLow dongles

ian77chen:
haLow: Add HaLow (802.11ah / S1G) support for USB HaLow dongles

Bring up the MM8108 USB HaLow (802.11ah) dongle on EAP115 with the 25.12 native netifd (no downgrade), including cloud configuration, telemetry and iwinfo support.

Driver and userspace
- morse_driver 1.17.8 with USB support, firmware, EAP115 board config and a 6.18 compat layer; hostapd_s1g / wpa_supplicant_s1g / morsecli bumped to 1.17.8.
- Fix an RCU self-detected stall and whole-system hang caused by an skbq lock deadlock on tx-status, and drop BAR frames which crash the MM8108 firmware.
- Handle the regulatory domain correctly on a country change: the driver country is a module parameter fixed at insertion time, so reload the module and avoid a global regulatory set that would clobber the 2.4/5G radios.

netifd integration
- Fix HaLow radio bring-up on native netifd: handler_load() read a wireless handler dump line by line, but the dump is a single newline-less JSON line and the morse handler forks a helper that inherits the dump fd, so the fd never reached EOF and the handler was silently dropped. Read the dump in one go instead. This also makes handler registration immune to fd inheritance from any shell handler.
- Guard against a null device in config_init so an unregistered handler no longer takes the 2.4/5G radios down with it.
- Derive the radio path for the USB dongle from sysfs so the capabilities lookup and the configuration apply can bind the radio.
- Add the missing dump-phase callbacks to the morse wireless handler.
- Emit qos_map_set only when Interworking/HS20 is enabled; the morse driver does not support it and an unconditional set failed AP bring-up.
- Default dtim_period to 1 for mesh, which mesh peers require to connect.

Interface naming
- Name the HaLow netdevs after the mac80211 convention used by the 2.4/5G radios: phyN-ap0 / phyN-sta0 / phyN-mesh0, with the phy resolved dynamically since the dongle may enumerate as any phyN. The mesh daemon and the schema no longer force a fixed interface name.

Break a layer-2 loop in WDS AP mode
- In WDS 4-addr AP mode hostapd creates a dynamic AP_VLAN child bridged next to the AP master port on the same radio, so a station's upstream broadcast was flooded cross-port and re-transmitted over the air back to the originating station. Isolate both bridge ports (a Linux bridge only blocks traffic between two ports when both are isolated) from a hotplug hook that re-fires on every association. Verified with 0% packet loss.

Cloud configuration and telemetry
- Recognize the HaLow band: scan for Morse PHYs even when board.json declares a wlan section (a USB dongle is not in board.json), and skip the HT mode match for HaLow, whose capabilities carry no HT/VHT/HE mode.
- Render channel, operating class and centre frequency from a per-country S1G channel map. These are per-country table values and the generic 5 GHz algorithm filtered out legal channels and emitted a wrong operating class.
- Resolve the HaLow country by following the 2.4/5G peer radio when HaLow has none, normalizing European codes to EU and falling back to US when a country is outside the supported set.
- Attribute survey data per interface. The HaLow phy operates on borrowed 5 GHz shim frequencies that collide with the real 5 GHz radio, so a frequency-only match cross-contaminated both radios' telemetry.
- Keep the HaLow radio in telemetry when a mesh and an AP interface coexist.
- Keep the capabilities in sync with the dongle by regenerating them from a hotplug event.

iwinfo
- Report HaLow the way the dot11ah backend does: channel in MHz, 802.11ah, the Morse hardware string, the real 802.11ah bitrate instead of the borrowed 5 GHz rate, WDS stations merged into the master interface, and a noise floor that is always a valid negative dBm.

Verified on device: HaLow AP and mesh come up as phy2-ap0 and phy2-mesh0, stations associate, mesh peers reach ESTAB, the data path works with no packet loss, and the 2.4/5G radios, Ethernet and cloud management are unaffected.

Fixes: WIFI-15595


tanyasingh-ec (changes requested):
modify commit msg: 
- the commit msg should start with lower case: "halow: add HaLow (802.11ah/S1G) support for USB HaLow dongles"
- remove '(wlan-ap part)'
- add support for EAP115a

venkatchimata (approved):

feeds/morse/kernel/mm-board-config/Makefile
SebastianHuang-ec, line 1: The board config version is supposed to match the firmware version. Also, the bcf_binaries repo is archived. Please refer to the upstream morse-feed to update the package source.

feeds/morse/kernel/morse-fw/Makefile
SebastianHuang-ec, line 1: Please use the upstream repo as the package source instead of uploading the binaries directly.

feeds/morse/netifd-morse/etc/init.d/halow-capabilities-boot-sync
```diff
@@ -0,0 +1,34 @@
...
+# window capabilities.uc's scan_morse_phys() sees no morse PHY and
+# writes a capabilities.json without HaLow, even though the dongle is
+# plugged. The 1-minute cron eventually catches this, but a device
+# owner watching the console during first boot sees a stale file for
+# up to 60s.
+#
+# This service fires ONCE, ~20s after boot completes (long enough for
+# firmware/debugfs to settle, well before the first cron tick), by
+# calling halow-capabilities-sync. If everything is already in sync
+# the sync script exits immediately (cheap). [ECCLOUD-10057]
```
SebastianHuang-ec: Please remove comments which include private information.

feeds/morse/netifd-morse/etc/uci-defaults/99-halow-capabilities-sync
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+#
+# Install a per-minute cron entry that runs halow-capabilities-sync on
+# EAP115/EAP115a. Catches boot-time races where capabilities.json ended
+# up out of sync with the actual HaLow dongle state. [ECCLOUD-10057]
```
SebastianHuang-ec: Please remove the comments which include private information.

feeds/morse/netifd-morse/lib/hotplug.d/ieee80211/30-morse-detect
```diff
@@ -0,0 +1,128 @@
...
+			uci -q set wireless.radio${devidx}.disabled=0
+			morse_committed=1
+			logger -t morse-hotplug "Enabled wireless.radio${devidx} (morse)"
+		fi
+		morse_radios="${morse_radios:+$morse_radios }radio${devidx}"
+	fi
+	devidx=$((devidx + 1))
+done
+
+# Refresh capabilities.json. [ECCLOUD-10057]
```
SebastianHuang-ec: Please remove the comments which include private information.

feeds/morse/netifd-morse/lib/hotplug.d/ieee80211/40-morse-remove
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
+#
+# Handle Morse Micro HaLow PHY removal on USB hotplug (dongle unplug).
+# Triggered when an ieee80211 PHY is unregistered. [ECCLOUD-10057]
```
SebastianHuang-ec: Please remove the comments which include private information.

feeds/morse/netifd-morse/lib/netifd/morse/morse_overrides.sh
```diff
@@ -818,6 +826,10 @@ morse_override_wpa_supplicant_add_network() {
 
 	[ "$_w_mode" = "mesh" ] && {
 		json_get_vars mesh_id dtim_period encryption
+		# ECCLOUD-9794: HaLow mesh cannot connect when the cloud config
```
SebastianHuang-ec: Please remove the comments which include private information.

SebastianHuang-ec (changes requested):

tanyasingh-ec:
@ian77chen  Since changes have been force pushed after @venkatchimata 's approval. Please make sure to let the reviewers know what changes you made and why

ian77chen:
> @ian77chen Since changes have been force pushed after @venkatchimata 's approval. Please make sure to let the reviewers know what changes you made and why

OK

ian77chen:
Hi @venkatchimata, I just modify feeds/morse/netifd-morse/etc/init.d/halow-gpio-reset.
Issue:
EAP115 hangs during boot when powered through its own USB port. Console
stops after "procd: - init -" and never reaches a login prompt.
Root cause:
halow-gpio-reset unconditionally wrote 0x44 to reg 0x01 of the I2C GPIO
expander at 0x41 to power the HaLow dongle. When the board is fed through
that same USB port, opening the gate back-feeds our own supply and wedges
boot, so no later init script runs.


Solution:
Gate the write on the power source. Reg 0x00 bit1 = VBUS present on the
connector, bit6 = we are sourcing it, so bit1 set with bit6 clear means
something external is feeding us - skip the write. Observed: 0x01 PoE
only, 0x02 USB-powered, 0x42 PoE with dongle powered. An unreadable
register also skips, since an unpowered dongle beats a hung boot.


Unit test:
Boots normally on PoE, on USB power, and on PoE with the HaLow dongle
attached.

feeds/morse/utils/morse_mesh11sd/patches/003-fix_ifname_confliction.patch
SebastianHuang-ec, line 1: Please rename the patch because this patch is not vendor-specific.

SebastianHuang-ec (changes requested):

SebastianHuang-ec (approved):

patches-25.12/0186-netifd-handler_load-EOF-guard-for-shell-handlers-that-leak-fd.patch
```diff
@@ -0,0 +1,75 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
tanyasingh-ec: use real revision hash

tanyasingh-ec (approved):

PR #1220, opened 2026-07-30
mediatek: BLE support for Edgecore EAP115

harrisonyeh:
Add Bluetooth Low Energy support for the EAP115 platform.

Features :

  - Zephyr base hci_uart firmware support for nordic nrf54l15
  - Bluez relative tools support
  - Bluetooth relative modules in linux kernel support
  - Initialization scripts to generate standard HCI interface
  - Simple script to generate ibeacon
  - Using a independent ble-support.yml to include bluez relative tools

Test results: https://telecominfraproject.atlassian.net/browse/WIFI-15598?focusedCommentId=64499

Fixes: WIFI-15598


SebastianHuang-ec (approved):

tanyasingh-ec (approved):

tanyasingh-ec:
@venkatchimata Please take a look when you get a chance

patches-25.12/0186-Add-uart1-uart2-and-pins-control-definition-for-eap115-board.patch
```diff
@@ -0,0 +1,49 @@
...
+ 			groups = "uart1_2";
+ 		};
+ 	};
++
++	uart2_pins: uart2-pins {
++		mux {
++			function = "uart";
++			groups = "uart2_1";
++		};
++	};
```
venkatchimata, line 25: This change is not just specific to EAP115. Does it impact any other platforms?

feeds/ble-support/firmware/nrf54l15/files/ble.init
```diff
@@ -0,0 +1,63 @@
+#!/bin/sh /etc/rc.common
+# Copyright (C) 2007 OpenWrt.org
+
```
venkatchimata, line 3: As this is a a common init script. Does it cause any issues for any other platforms other than EAP115?

patches-25.12/0186-Add-uart1-uart2-and-pins-control-definition-for-eap115-board.patch
harrisonyeh: The ibeacon is sent through bluetoothctl (bluez tool). 
It only works on devices which is bluez supported.
The future BLE framework will provide full support including both BlueZ-supported and non-BlueZ-supported devices.
So, in the next release, we should remove the old version and replace it with the BLE framework.

feeds/ble-support/firmware/nrf54l15/files/ble.init
harrisonyeh: The ibeacon is sent through bluetoothctl (bluez tool). It only works on devices which is bluez supported. The future BLE framework will provide full support including both BlueZ-supported and non-BlueZ-supported devices. So, in the next release, we should remove the old version and replace it with the BLE framework.

patches-25.12/0186-Add-uart1-uart2-and-pins-control-definition-for-eap115-board.patch
harrisonyeh: This definition exists in mt7987-pinctrl.dtsi. Although some development boards reference this file, no development board design uses UART2; currently, only the EAP115 uses it as its BLE HCI interface.

mkwacz (approved):

PR #1223, opened 2026-07-31
patches-25.12: increase ramoops size for EAP111 and EAP112

SebastianHuang-ec:
Increase the ramoops size to 512KiB

Fixes: WIFI-15307


venkatchimata (approved):

tanyasingh-ec (approved):
LTGM; will merge after build OK

PR #1229, opened 2026-08-05
ucentral:add -i flag for QA certificates to fix connection failure

lyc-cybertan:
WIFI-15602
Bug Phenomenon
When using QA certificates to connect to uCentral, the connection fails with OpenSSL error X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN (19). This error indicates that during certificate chain construction, a self-signed certificate was encountered that is not present in the local trusted CA certificate store.

In v4.2.1, the -i parameter was always added when using QA certificates, allowing the connection to succeed. 
However, in v5.0.0, the following change was introduced:

[ "$insecure" = "true" ] && insecure=1 || insecure=0

This logic fixed insecure mode being enabled by default when allow-self-signed is unset, but broke QA certificate connections.

Fix
Add a conditional check to detect QA certificates and explicitly add the -i parameter when a QA certificate is being used.

Fixes: WIFI-15602-QA certificates are unable to connect to controller

Testing
- Verified that QA certificate connections work correctly with the fix
  applied
- Confirmed that production certificates continue to work without the
  -i parameter

### DIFF
--- a/feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
+++ b/feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
@@ -44,6 +44,12 @@ start_service() {
 		[ "${selfsigned}" == "true" ] && insecure=1
 	fi
 
+	if [ -f /etc/ucentral/cert.pem ]; then
+		case "$(openssl x509 -in /etc/ucentral/cert.pem -noout -issuer 2>/dev/null)" in
+			*"OpenLAN Demo"*) insecure=1 ;;
+	esac
+	fi
+
 	server=
 	[ -f /etc/ucentral/gateway.json ] && server=$(cat /etc/ucentral/gateway.json | jsonfilter -e '@["server"]')
 	port=


feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
```diff
@@ -44,6 +44,12 @@ start_service() {
 		[ "${selfsigned}" == "true" ] && insecure=1
 	fi
 
+	if [ -f /etc/ucentral/cert.pem ]; then
+		case "$(openssl x509 -in /etc/ucentral/cert.pem -noout -issuer 2>/dev/null)" in
+		*"OpenLAN Demo"*) insecure=1 ;;
+	esac
```
venkatchimata, line 50: esac is not aligned correctly. Otherwise the changes look good tome. I want other experts too to review this and check if all is good.

lyc-cybertan: @venkatchimata Thank you for your suggestion. I have made the correction

lyc-cybertan: You can view more detailed context at https://telecominfraproject.atlassian.net/browse/WIFI-15602?focusedCommentId=61905.

When we use the QA certificate to execute the following command to connect to the server:
" openssl s_client -connect gw-qa01.lab.open-lan.org:15002 \
 -cert /etc/ucentral/operational.pem -key /etc/ucentral/key.pem \
 -CAfile /etc/ucentral/operational.ca -servername gw-qa01.lab.open-lan.org </dev/null "
We observed the appearance of "Verify return code: 19 (self-signed certificate in certificate chain)" in the returned results.
The top of the certificate chain issued by the server is a self-signed root certificate (i.e., OpenLAN Root CA). Since the client cannot find a matching trusted entry in the specified -CAfile (/etc/ucentral/operational.ca), it fails to establish a complete trust chain and treats the root certificate as an "unknown self-signed certificate," resulting in the X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN error.
Reviewing the relevant client code revealed that when the `-i` parameter is enabled, it activates the flag `LCCSCF_ALLOW_SELFSIGNED`. The underlying logic of this flag is to explicitly instruct the client to ignore the specific error `X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN`. Therefore, even if the verification code is 19, the client will still allow the TLS connection to proceed (i.e., skip the strict validation of the self-signed root certificate), ensuring that business communication can continue normally in test environments or internal networks even if the certificate chain is incomplete

venkatchimata (approved):

feeds/ucentral/ucentral-client/files/etc/init.d/ucentral
tanyasingh-ec: @lyc-cybertan is the alignment done? I can't see the difference 🤔

lyc-cybertan: @tanyasingh-ec  Yes, on the previous line 49, the format is as follows
case "$(openssl x509 -in /etc/ucentral/cert.pem -noout -issuer 2>/dev/null)" in
*"OpenLAN Demo"*) insecure=1 ;;
esac
Now updated

Image

tanyasingh-ec (approved):

PR #1231, opened 2026-08-05
ath12k: fix peer collision when a client moves between VAPs on one radio

tanyasingh-ec:
ath12k allows one peer per MAC per pdev (radio), so ath12k_peer_create() rejects with -EINVAL ("Peer already found in pdev_idx.. vdev..") when a stable-MAC client (re)associates to another VAP on the same radio before its previous peer is torn down. hostapd then loops on "Could not add STA to kernel driver". Seen on EAP105 (up to 8 VAPs/band).

ath12k counterpart of 1d106efd ("ath11k: fix multi band roaming", WIFI-12188), extended to the same-pdev case: on the cross-vdev, non-MLO collision, delete the stale peer on its original vdev (real WMI delete, since the firmware enforces one peer per pdev) and continue with the create. Same-vdev/MLO keep the original reject.

Fixes: WIFI-14695


venkatchimata (approved):

SebastianHuang-ec (approved):

tanyasingh-ec:
waiting for @ArifAlam 's feedback before merge

ArifAlam (approved):

PR #1233, opened 2026-08-05
rrmd: scan: restrict off-channel scan to the target frequency

mkwacz:
The periodic RRM scan set wiphy_freq/center_freq1 on the TRIGGER_SCAN request but never populated scan_frequencies, so nl80211 fell back to scanning the whole band on every off-channel dwell.

Limit the request to the single target frequency via scan_frequencies and drop the active scan_ssids probe, so the off-channel visit is a short passive dwell bounded by the existing measurement_duration.

Fixes: WIFI-14822

### DIFF
--- a/feeds/ucentral/rrmd/files/usr/share/rrmd/scan.uc
+++ b/feeds/ucentral/rrmd/files/usr/share/rrmd/scan.uc
@@ -12,7 +12,7 @@ let scan_blocked_cnt = 0;
 function scan(phy, params) {
 	if (params.wiphy_freq) {
 		params.center_freq1 = (int) (params.wiphy_freq) + (int) (phys[phy].offset);
-		params.scan_ssids = [ '' ];
+		params.scan_frequencies = [ (int) (params.wiphy_freq) ];
 	}
 
 	params.scan_flags = SCAN_FLAG_AP;


venkatchimata (approved):

mkwacz:
Validated on CI image -> passed

On the TIP base rrmd, scanning is off out of the box (ubus call rrm scan_dump empty, no periodic scan). To enable periodic channel survey and off-channel neighbor scanning, uncomment three lines and restart rrmd:

    /usr/share/rrmd/local.uc - in init(), uncomment the periodic survey timer:
    uloop_timeout(interfaces_update, 5000);

    /usr/share/rrmd/scan.uc - at the end of scan_timer(), uncomment the re-arm and use 5 s (1000 ms is too aggressive):
    return 5000;

    /usr/share/rrmd/scan.uc - in nl_handle_scan_msg(), uncomment the scan-result fetch/parse:
    get_scan_res(msg.msg?.dev);

Then restart: /etc/init.d/rrmd restart

Verify:

    ubus call rrm scan_dump → neighbor list with ssid/freq/signal (the map is keyed by BSSID, so grep ssid/signal, not bssid).

    ubus call rrm interfaces → per-BSS load and noise.

PR #1242, opened 2026-08-12
mediatek: enable USB host and fix LTE GPIOs on Edgecore EAP112

ian77chen:
To support LTE functions on Edgecore EAP112.

The EAP112 has its LTE modem on the SoC USB port, but that port was never brought up: mt7981b.dtsi defaults &xhci and &usb_phy to disabled and the board DTS did not override them, and Device/edgecore_eap112 did not select kmod-usb3, so the MediaTek xHCI driver was not built.

The USB function drivers (qmi_wwan, cdc_mbim, option) still loaded, so nothing looked broken, but without a host controller there was no bus to enumerate on: /sys/bus/usb/devices stayed empty and the wwan proto handler always failed with `No valid device was found'.

The modem power and reset lines were unreachable as well. Their sysfs GPIO numbers date back to a kernel 5.4 tree, where gpiolib picked a chip base by counting down from ARCH_NR_GPIOS, placing the 57-pin mt7981 controller at 455 so that pio pins 2 and 3 showed up as 457 and 458. Current kernels count up from GPIO_DYNAMIC_BASE (512) instead, so the exports failed with invalid GPIO 457 and 458. Renumber them to 514 and 515, which are the same two pins.

Fixes: WIFI-15594

### DIFF
--- a/patches-25.12/0190-mediatek-enable-usb-host-for-Edgecore-EAP112.patch
+++ b/patches-25.12/0190-mediatek-enable-usb-host-for-Edgecore-EAP112.patch
@@ -1,0 +1,77 @@
+From d4e5f60718293a4b5c6d7e8f90123456789abcde Mon Sep 17 00:00:00 2001
+From: Ian Chen <ian77_chen@accton.com>
+Date: Fri, 7 Aug 2026 15:30:00 +0800
+Subject: [PATCH] mediatek: enable USB host and fix LTE GPIOs on Edgecore EAP112
+
+The EAP112 has its LTE modem on the SoC USB port, but that port was
+never brought up: mt7981b.dtsi defaults &xhci and &usb_phy to disabled
+and the board DTS did not override them, and Device/edgecore_eap112 did
+not select kmod-usb3, so the MediaTek xHCI driver was not built.
+
+The USB function drivers (qmi_wwan, cdc_mbim, option) still loaded, so
+nothing looked broken, but without a host controller there was no bus to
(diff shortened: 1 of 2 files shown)


patches-25.12/0190-mediatek-enable-usb-host-for-Edgecore-EAP112.patch
```diff
@@ -0,0 +1,77 @@
...
+ 	ucidef_add_gpio_switch "modem_power" "Modem power" "modem_power" "0"
+diff --git a/target/linux/mediatek/image/filogic.mk b/target/linux/mediatek/image/filogic.mk
+--- a/target/linux/mediatek/image/filogic.mk
++++ b/target/linux/mediatek/image/filogic.mk
+@@ -1388,7 +1388,7 @@ define Device/edgecore_eap112
+   IMAGES := sysupgrade.bin factory.bin
+   IMAGE/factory.bin := append-ubi | check-size $$$$(IMAGE_SIZE)
+   IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
+-  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
++  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware kmod-usb3
```
SebastianHuang-ec, line 71: I think that it is better to include "kmod-usb3" at only one place, here or in profile.

PR #1244, opened 2026-08-13
luci-mod-ucentral: fix uplink config not applied in Maverick mode

Javen829:
Problem:
The uplink config is not set in the /etc/config/network in Maverick mode.

Root Cause:
If /etc/ucentral/ucentral.active exists,
/usr/share/ucentral/ucentral.uc /etc/ucentral/ucentral.cfg.0000000001 is not executed. As a result, after configuring PPPoE in Maverick mode, ucentral does not read /etc/ucentral/profile.json and does not update the UCI configuration.

Solution:
After configuring the network in Maverick mode, profileupdate is triggered. Add a step to remove the symbolic link so that the UCI configuration can be updated after the AP reboots.

Tests: Successfully verified on RAP7110C-341X

Fixes: WIFI-15336

### DIFF
--- a/feeds/tip/luci/luci-mod-ucentral/root/sbin/profileupdate
+++ b/feeds/tip/luci/luci-mod-ucentral/root/sbin/profileupdate
@@ -11,5 +11,8 @@ else
 	/etc/init.d/firstcontact enable 
 	/etc/init.d/ucentral disable    
 fi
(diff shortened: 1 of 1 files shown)


tanyasingh-ec (approved):

jemmy7780 (approved):

mkwacz (approved):

jemmy7780:
@tanyasingh-ec  
       could you help merge this PR?

PR #1248, opened 2026-08-13
mtk: fix RAP630W-211G Factory partition corruption breaks wireless

Niko525620:
Root cause:
The Factory partition stores RF calibration data and is read-only. If this partition gets corrupted due to something other than a software issue, the wireless interface won't start properly.

Solution:
Store the backup RF calibration file at
/lib/firmware/mediatek/mt7981_eeprom_mt7976_dbdc.bin to avoid the above‑mentioned issue.

Fixes: WIFI-15655


tanyasingh-ec (approved):

feeds/mediatek/mt76/Makefile
```diff
@@ -633,6 +633,11 @@ ifeq ($(CONFIG_TARGET_PROFILE), "DEVICE_edgecore_eap111")
 	cp ./edgecore_eap111_eeprom_dbdc.bin \
 		$(1)/lib/firmware/mediatek/mt7981_eeprom_mt7976_dbdc.bin
 endif
+
+ifeq ($(CONFIG_TARGET_PROFILE), "DEVICE_sonicfi_rap630w_211g")
+	cp ./sonicfi_rap630w_211g_eeprom_dbdc.bin \
+		$(1)/lib/firmware/mediatek/mt7981_eeprom_mt7976_dbdc.bin
+endif
```
venkatchimata, line 640: The changes look ok and may work too. Can this be improved instead of copying model specific files like this. 
cc: @tanyasingh-ec

Niko525620: Hi @venkatchimata 
Before the device leaves the factory, the RF calibration data is written into the Factory partition, and users cannot manually modify it afterwards.  Replacing the NAND flash without redoing the RF calibration can lead to this situation.  Additionally, bad blocks in the NAND flash can also cause similar issues.  The current solution is to prevent this special situation from occurring.

tanyasingh-ec: @venkatchimata The current design per model makes sense as the bdf per model differs.

venkatchimata (approved):

tanyasingh-ec:
@Niko525620  follow guidelines and fix the commit title

Niko525620:
> @Niko525620 follow guidelines and fix the commit title

@tanyasingh-ec I have completed the modification. Please check it again. Thank you

PR #1250, opened 2026-08-13
patches-25.12: fix led default brightness being cleared by the trigge…

tanyasingh-ec:
…r write

ucidef_set_led_default() has no effect: load_led() in /etc/init.d/led applied the configured brightness before setting the trigger, and writing 'none' to trigger reaches led_trigger_remove() in the kernel LED core, which calls led_set_brightness(led_cdev, LED_OFF) and zeroes the value just written. Since ucidef_set_led_default() sets no trigger, load_led() falls back to 'none' and the LED is switched off every time.

This affects every target that declares a solid-on indicator this way, 10 entries in mediatek/filogic 01_leds alone. It went unnoticed because most such LEDs sit on the led-running GPIO, where /etc/diag.sh set_state done or ucentral-state lights them independently; only boards without a second owner are visibly dark.

Add patch 0190, which moves the default-brightness block below the trigger write. It stays above the $ret check so a default is still applied when the trigger is unavailable for want of a kernel module, leaving the LED plain on rather than off.

The preceding commit gives EAP111, EAP112 and EAP115 a power LED via ucidef_set_led_default, so none of them would light without this.

Fixes: WIFI-15656

### DIFF
--- a/patches-25.12/0190-base-files-led-apply-default-brightness-after-setting.patch
+++ b/patches-25.12/0190-base-files-led-apply-default-brightness-after-setting.patch
@@ -1,0 +1,58 @@
+From a299b1c784252a43f094dd2ee1f0536b910b5ccc Mon Sep 17 00:00:00 2001
+From: Tanya Singh <tanya_singh@accton.com>
+Date: Thu, 13 Aug 2026 19:41:40 +0800
+Subject: [PATCH] base-files: led: apply the led default brightness after
+ setting the trigger
+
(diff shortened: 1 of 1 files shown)


venkatchimata (approved):

tanyasingh-ec:
@SebastianHuang-ec suggested moving both default-brightness branches below the trigger write so the
value sticks.

venkatchimata (approved):

SebastianHuang-ec (approved):

PR #1254, opened 2026-08-13
certificates: mount_certs: skip ubiattach -d 3 if ubi0 is attached

HenryHaller:
Commit b4e1583fa (WIFI-15591) added early-boot certificate restoration. On single-NAND IPQ53xx targets (e.g. cig_wf189, cig_wf189h), partition 'fs' / 'rootfs_1' is attached by U-Boot as ubi0. Calling ubiattach -m $MTD -d 3 forces the kernel to attach the primary UBI partition as a secondary device, corrupting the UBI volume table (dropping 'kernel' and 'ubi_rootfs' volumes) and dropping the AP into U-Boot on reboot.

Check if /dev/ubi0 exists before calling ubiattach -d 3. If ubi0 is already attached, mount_certs returns safely without mutating primary UBI headers.

Fixes: WIFI-15659

### DIFF
--- a/feeds/tip/certificates/files/usr/bin/mount_certs
+++ b/feeds/tip/certificates/files/usr/bin/mount_certs
@@ -131,13 +131,21 @@ udaya,a6-od2)
 		PART_NAME=rootfs_1
 	fi	
 	;;
-cig,wf186w|\
 cig,wf189|\
+cig,wf189h)
+	if grep -q rootfs_1 /proc/cmdline || grep -q "ubi.mtd=rootfs_1" /proc/cmdline; then
+		PART_NAME=rootfs
+	else
+		PART_NAME=rootfs_1
+	fi
+	;;
+cig,wf186w|\
 cig,wf189w|\
-cig,wf189h|\
 cig,wf186h|\
 cig,wf196|\
-cig,wf188n|\
+cig,wf188n)
+	PART_NAME=rootfs_1
+	;;
 emplus,wap380c|\
 emplus,wap385c|\
 emplus,wap386v2|\
@@ -170,6 +178,9 @@ esac
 MTD=$(find_mtd_index $PART_NAME)
 
 [ -z "$MTD" ] && return 1
+
+# Skip secondary ubiattach if partition is already attached as ubi0
+[ -e /sys/class/ubi/ubi0/mtd_num ] && [ "$MTD" = "$(cat /sys/class/ubi/ubi0/mtd_num)" ] && return 0
 
 ubiattach -m $MTD -d 3
 if [ -e /dev/ubi3 ] ; then


feeds/tip/certificates/files/usr/bin/mount_certs
```diff
@@ -171,6 +171,9 @@ MTD=$(find_mtd_index $PART_NAME)
 
 [ -z "$MTD" ] && return 1
 
+# Skip secondary ubiattach if partition is already attached as ubi0
+[ -e /dev/ubi0 ] && return 0
```
phwhite: This is a breaking change for PKI.  /dev/ubi0 will pretty much always be mounted at boot time, so my understanding of this is that it basically is like adding an exit 0 here.

The main intentions of this code (which must be here for the script to work properly) is:
* Check if /certificates/ on current UBI (ubi0) mounted (rootfs_1 or rootfs_2) contains certificattes
* If it doesn't, attempt to attach the opposite bank (opposite of current rootfs_1 or rootfs_2) to ubi3 and see if it has the certificates

This logic only kicks-in on every-other-upgrade where boot-bank is swapped.  You may not notice a breaking change right away.

I don't see how this could attach the same MTD partition twice, trying on my AP I get a attach failure E_BUSY.  But if somehow that is what's happening, then the issue here would be that it fell through to this part of the code with $MTD being the currently mounted rootfs_1 or rootfs_2 instead of the opposite boot bank.

This is my quick read on this, needs someone else with more time to review it deeper.

phwhite (changes requested):

feeds/tip/certificates/files/usr/bin/mount_certs
phwhite, line 156: @HenryHaller I think the issue is here.   This is always defaulting to `rootfs_1` without checking what it actually should be.  See line 128 above how other platforms check which PART to define here.

phwhite: I would recommend a eng from CIG take a look at this, too.  @jaspreetsachdev can you help tag someone from CIG? I'm not sure who works on the github stuff.   Thank you

feeds/tip/certificates/files/usr/bin/mount_certs
HenryHaller: Thank you @phwhite for your feedback and insights,

I have updated the commit to be much more selective about when it returns early, specifically checking whether `$MTD` matches the currently active UBI partition (`/sys/class/ubi/ubi0/mtd_num`):

```sh
[ -e /sys/class/ubi/ubi0/mtd_num ] && [ "$MTD" = "$(cat /sys/class/ubi/ubi0/mtd_num)" ] && return 0
```

This ensures dual-bank A/B upgrades can still attach the opposite bank to `ubi3` as intended, while preventing single-bank / same-bank targets from attempting `ubiattach` on the active `$MTD`.

That said, I understand that any change to core OpenWiFi early boot code requires thorough review across multiple hardware targets. Fortunately, based on further testing, it seems like the production impact of this issue is less critical than I initially thought, so I will likely withdraw this PR. We are testing using this refined check in our downstream overlay builds to eliminate a non-critical `EBUSY` / `EEXIST` mounting conflict on boot.

feeds/tip/certificates/files/usr/bin/mount_certs
```diff
@@ -145,7 +145,11 @@ emplus,wap581|\
 asterfusion,AP7330|\
 yuncore,ax840|\
 yuncore,fap655)
-	PART_NAME=rootfs_1
+	if grep -q rootfs_1 /proc/cmdline || grep -q "ubi.mtd=rootfs_1" /proc/cmdline; then
```
phwhite: @HenryHaller I would recommend splitting this out just for the wf189 models, until you confirm each model in this case list support dual-boot.   But the ODMs can help confirm

phwhite: @HenryHaller But this approach looks better, and I'm hopeful it will fix your problem!

phwhite: Oh @HenryHaller you already handled this!!  Looks good

feeds/tip/certificates/files/usr/bin/mount_certs
HenryHaller: I changed around the code in here a bit to address this.

PR #1281, opened 2026-09-01
MPSKD: fix vlan assignment 

mkwacz:
hostapd, mpskd: fix VLAN assignment lost on MPSK roaming between bands
When a client roams between bands (2.4/5/6 GHz) using RADIUS-based
Multi-PSK (MPSK), the VLAN assignment is lost and the client gets
placed on the default VLAN 1.

Fixes: WIFI-14179

### DIFF
--- a/feeds/qca-wifi-7/hostapd/files/mpskd
+++ b/feeds/qca-wifi-7/hostapd/files/mpskd
@@ -252,14 +252,34 @@ function sta_auth_psk(ifname, addr) {
 	return ssid_psk(ssid, addr);
 }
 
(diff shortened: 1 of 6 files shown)


feeds/qca-wifi-7/hostapd/files/mpskd
venkatchimata, line 332: Do we need any null check for data?

mkwacz: It is not a part of my commit.

venkatchimata (approved):

PR #1282, opened 2026-09-04
filogic: enable WED offload on RAP630W-211G

lyc-cybertan:
The Sonicfi RAP630W-211G (MT7981B + MT7976) probes mt7915e without wed_enable, so the Wireless Ethernet Dispatch offload path is never initialised and the bridged LAN->WLAN path is forwarded by the CPU. 5G RvR LAN-to-WiFi throughput does not meet spec.

Setting wed_enable=1: the frames of a hardware offloaded flow never traverse the tc qdisc of the wireless netdev, so an SSID carrying a rate_limit silently stops being shaped. WED is therefore turned on here together with the two changes that keep the rate limiter working.

Three patches, in the order they take effect:

1. ucentral-schema 0001-renderer-*: report capa.wed when a loaded wireless module was probed with wed_enable=1, so the renderer keys off the platform really having the offload engine rather than off a board name. Default bridger to 'no-restart' instead of stopped when capab.wed is set; set_enabled() keeps the first state written, so an SSID with strict_forwarding still promotes bridger to 'early' and a platform without WED is unchanged. Give a rate-limited SSID an explicit wlanr<n> ifname and add it to bridger.@defaults[0].blacklist, the same exclusion captive portal interfaces already use.

2. bridger 002-bridger-fix-mtk-*: bridger only consults the blacklist for the device it attaches the BPF classifier to. A flow is learned from its ingress device, so a flow whose egress device is blacklisted was still given a tc-flower rule with SKIP_SW and redirected into that device purely in hardware, bypassing the HTB qdisc that enforces the per-SSID egress rate. Skip only the hardware binding, so a blacklisted egress device is still bridged by bridger's BPF fast path.

3. patches-25.12/0123-mediatek-mt76-*: set MODPARAMS.mt7915e:=wed_enable=1 for the DEVICE_sonicfi_rap630w_211g profile only, so every other mt7915e board keeps the upstream default of WED disabled.

Fixes: WIFI-15633

RVR test:
before add ucentral-schema 0001-renderer-* and patches-25.12/0123-mediatek-mt76-*
image
already add:
image

Sanity test:
before add those three patches
image
already add:
image

### DIFF
--- a/feeds/ucentral/bridger/patches/001-renderer-enable-bridger-hardware-offload-on-WED-capa.patch
+++ b/feeds/ucentral/bridger/patches/001-renderer-enable-bridger-hardware-offload-on-WED-capa.patch
@@ -1,0 +1,92 @@
+diff -Nur ucentral-schema-2026.08.26~d1e90a04-orig/renderer/templates/interface/ssid.uc ucentral-schema-2026.08.26~d1e90a04/renderer/templates/interface/ssid.uc
+--- ucentral-schema-2026.08.26~d1e90a04-orig/renderer/templates/interface/ssid.uc	2026-09-03 14:29:15.338389105 +0800
++++ ucentral-schema-2026.08.26~d1e90a04/renderer/templates/interface/ssid.uc	2026-09-03 14:39:10.167216528 +0800
+@@ -85,6 +85,10 @@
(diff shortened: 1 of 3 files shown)


tanyasingh-ec:
@lyc-cybertan please move the test results to the Jira ticket.

jaspreetsachdev:
@lyc-cybertan why is this closed? @JackTsai-CBT

PR #1283, opened 2026-09-07
qca-wifi-6: fix uneven CPU load from packet steering on ipq50xx

crychic-cybertan:
Root cause:
QCA IPQ50xx platforms (qca-wifi-6) ship with network.globals.packet_steering enabled by default. Hardware acceleration (PPE/NSS/SFE) already handles packet distribution on these platforms, so software RPS (packet_steering) provides no benefit and instead causes uneven CPU load distribution across cores during throughput testing.

Solution:
Add a uci-defaults script that sets network.globals.packet_steering and network.globals.steering_flows to '0' by default. The defaults are only applied when not already configured, so any existing user settings are preserved across upgrades.

Fixes: WIFI-14504

### DIFF
--- a/feeds/qca-wifi-6/ipq50xx/base-files/etc/uci-defaults/35-packet-steering
+++ b/feeds/qca-wifi-6/ipq50xx/base-files/etc/uci-defaults/35-packet-steering
@@ -1,0 +1,21 @@
+#!/bin/sh
+# QCA IPQ platforms: default packet_steering to disabled.
+# Hardware acceleration (PPE/NSS/SFE) handles packet distribution;
+# software RPS provides no benefit and may cause uneven CPU load.
+# Only set defaults if not already configured (preserves user settings).
+
+changed=0
+
+if [ -z "$(uci -q get network.globals.packet_steering)" ]; then
+    uci set network.globals.packet_steering='0'
+    changed=1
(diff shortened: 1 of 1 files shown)


venkatchimata (approved):

tanyasingh-ec (changes requested):
Minor suggestions:
1. Ensure file is executable on installation (chmod 0755 in packaging or set in feed metadata)
2. Add a trailing newline at EOF

crychic-cybertan:
> Minor suggestions:
> 
> 1. Ensure file is executable on installation (chmod 0755 in packaging or set in feed metadata)
> 2. Add a trailing newline at EOF

@tanyasingh-ec 
The script permissions have been confirmed as chmod 0755.
A trailing newline has been added at the end‑of‑file.

crychic-cybertan (commented):
Thanks for the review! Addressed both points:

1. File is already executable in git (`100755`), matching the qca-wifi-7 reference commit (f4a62f9). OpenWrt's build preserves source-tree file modes when packaging base-files, so no extra chmod is needed in the Makefile/feed metadata.
2. Added the missing trailing newline at EOF — pushed in ba8d4734.

tanyasingh-ec (approved):

PR #1287, opened 2026-09-14
mac80211: fix the kernel crash caused by desc_va

Niko525620:
Constantly going through the wireless network loop of connecting -> sending traffic -> disconnecting can possibly crash the kernel.

Root Cause:
- When extracting the physical address of the packet buffer from the hardware descriptor, there was no safety or validity check.
- If this address is a wild pointer, directly accessing it when the address is invalid could trigger a kernel panic.

Solution:
used the kernel's safe access function copy_from_kernel_nofault() to perform a safety validation on this address.

Fixes: WIFI-15687


feeds/qca-wifi-7/mac80211/src/drivers/net/wireless/ath/ath12k/wifi7/hal_rx.c
```diff
@@ -452,8 +453,25 @@ int ath12k_wifi7_hal_wbm_desc_parse_err(struct ath12k_dp *dp, void *desc,
...
+		 * instead of crashing later when rel_info->rx_desc is
+		 * dereferenced.
+		 */
+		if (desc_va &&
+		    copy_from_kernel_nofault(&val, &desc_info->magic, sizeof(val))) {
+			dp->device_stats.invalid_rbm++;
+			ath12k_warn(dp->ab,
+				    "invalid rx_desc VA 0x%llx from HW CC, dropping\n",
+				    desc_va);
+			return -EINVAL;
```
venkatchimata, line 471: 1. When desc_va is zero the whole expression is false, so it skips the error return and assigns rel_info->rx_desc = desc_info, which is NULL.
2. The magic value is read but never checked. It copies into val and then discards it. So the probe proves the address is mapped, not that it points at a descriptor.
3. dp->device_stats.invalid_rbm++ is reused for a bad VA, but that counter means invalid return buffer manager. Two distinct faults in one statistic makes later diagnosis harder; this deserves its own counter.
4. No rate limiting on the warn. If the condition repeats per packet, the log floods.
5. The \n on the ath12k_warn is inconsistent with the existing call a few lines above, which has none.
6. And it uses dp->ab where the function already has struct ath12k_base *ab in scope.

jemmy7780: @venkatchimata 
Thank you for your suggestions. Here are my comments:
1. No, I think there is no need to consider the case where desc_va is null, because the place where this function is called already has handling for a null desc_va. Please see the following code:
    backports-6.6.15/drivers/net/wireless/ath/ath12k/wifi7/dp_rx.c
    int ath12k_wifi7_dp_rx_process_wbm_err{
    ....
               ret = ath12k_wifi7_hal_wbm_desc_parse_err(dp, rx_desc,
                                                          &err_info);
                if (ret) {
                        DP_DEVICE_STATS_INC(dp, wbm_err.drop[WBM_ERR_DESC_PARSE_ERROR], 1);
                        ath12k_warn(ab,
                                    "failed to parse rx error in wbm_rel ring desc %d\n",
                                    ret);
                        continue;
                }

                desc_info = err_info.rx_desc;

                /* retry manual desc retrieval if hw cc is not done */
                if (!desc_info) {
                        DP_DEVICE_STATS_INC(dp, wbm_err.drop[WBM_ERR_GET_SW_DESC_FROM_CK_ERROR], 1);
                        desc_info = ath12k_dp_get_rx_desc(dp, err_info.cookie);
                        if (!desc_info) {
                                DP_DEVICE_STATS_INC(dp, wbm_err.drop[WBM_ERR_GET_SW_DESC_ERROR], 1);
                                ath12k_warn(ab, "Invalid cookie in DP WBM rx error descriptor retrieval: 0x%x\n",
                                            err_info.cookie);
                                continue;
                        }
                }

                if (desc_info->magic != ATH12K_DP_RX_DESC_MAGIC)
                        ath12k_warn(ab, "WBM RX err, Check HW CC implementation");

                msdu = desc_info->skb;
                desc_info->skb = NULL;

2. Other places where the magic is incorrect also just print a warn log and don't do any special handling.
3. we will remove ath12k_warn() to avoid this issue
4. Same as above
5. Same as above
6 .You're right, we will just use the struct point ab directly.

PR #1289, opened 2026-09-14
ath12k: resolve peer by vdev id in peer delete and pass the correct value of each TID during a stop AMPDU session

ArifAlam:
This PR fixes two ath12k issues related to peer cleanup and AMPDU session handling:

- Resolve peers by `(vdev_id, MAC address)` in the peer delete path instead of relying on the shared MAC-address hash table. This prevents a cross-band roam from causing the delete path to modify the peer belonging to another radio, which could corrupt peer-list entries and degrade the RX fast path.
- Pass the TID-specific RX state when stopping an AMPDU session. Previously, the base address of the RX TID array was passed, causing the operation to target TID 0 even when another TID requested the DELBA. This could incorrectly stop data-TID aggregation and reduce TCP throughput.

Fixes: WIFI-15461


venkatchimata (approved):

tanyasingh-ec:
@ArifAlam  Please follow PR guidelines

tanyasingh-ec (approved):

PR #1290, opened 2026-09-16
Mediatek:CVE-2026-31736

hehuimin-cybertan:
net: ethernet: mtk_ppe: avoid NULL deref when gmac0 is disabled If the gmac0 is disabled the precheck for a valid ingress device will cause a NULL pointer deref and crash the system. This happens because eth->netdev[0] will be NULL but the code will directly try to access netdev_ops. Instead of just checking for the first net_device, it must be checked if any of the mtk_eth net_devices is matching the netdev_ops of the ingress device.

Fixes: WIFI-15707


tanyasingh-ec (changes requested):
Please follow the PR guidelines, example: the branch naming is not correct

SebastianHuang-ec (approved):
The patch is OK

tanyasingh-ec (approved):

PR #1295, opened 2026-09-17
mac80211: fix the kernel crash caused by desc_va

Niko525620:
Constantly going through the wireless network loop of connecting -> sending traffic -> disconnecting can possibly crash the kernel.

Root Cause:
- When extracting the physical address of the packet buffer from the hardware descriptor, there was no safety or validity check.
- If this address is a wild pointer, directly accessing it when the address is invalid could trigger a kernel panic.

Solution:
used the kernel's safe access function copy_from_kernel_nofault() to perform a safety validation on this address.

Fixes: WIFI-15687


Niko525620:
@venkatchimata 
1.desc_info will be checked after returning, so no need to check it here
2.Remove the count variable and the warning log; let the upper-level caller handle validation and printing.

mkwacz (approved):

Niko525620:
@mkwacz 
The magic check is omitted here for the following reason:
1. The `ath12k_dp_rx_process_wbm_err` function in `backports-6.5-rc3/drivers/net/wireless/ath/ath12k/dp_rx.c` already implements magic validation. This function only prints a warning without performing any dedicated exception handling.
```
               ret = ath12k_hal_wbm_desc_parse_err(ab, rx_desc, &err_info);
                if (ret) {
                        scnprintf(buf, sizeof(buf), "failed to parse rx error in wbm_rel ring desc %d\n", ret);
                        ath12k_err_dump(ab, buf, "wbm err desc: ", rx_desc,
                                        sizeof(struct hal_wbm_release_ring), srng);
                        BUG_ON(1);
                }

                desc_info = (struct ath12k_rx_desc_info *)err_info.rx_desc;

                /* retry manual desc retrieval if hw cc is not done */
                if (!desc_info) {
                        desc_info = ath12k_dp_get_rx_desc(ab, err_info.cookie);
                        if (!desc_info) {
                                ath12k_warn(ab, "Invalid cookie in manual desc retrieval cookie 0x%x", err_info.cookie);
                                continue;
                        }
                }

                if (desc_info->magic != ATH12K_DP_RX_DESC_MAGIC)
                        ath12k_warn(ab, "WBM RX err, Check HW CC implementation");
```
2. Other places where the magic is incorrect also just print a warn log and don't do any special handling.

